PR #1, opened 2018-11-20
Add glances config

aptalca:

### DIFF
--- a/glances.subdomain.conf.sample
+++ b/glances.subdomain.conf.sample
@@ -1,0 +1,29 @@
+# make sure that your dns has a cname set for glances and that your glances container is not using a base url
+
+server {
+    listen 443 ssl;
+
+    server_name glances.*;
+
+    include /config/nginx/ssl.conf;
+
+    client_max_body_size 0;
+
+    # enable for ldap auth, fill in ldap details in ldap.conf
+    #include /config/nginx/ldap.conf;
+
+    location / {
+        # enable the next two lines for http auth
+        #auth_basic "Restricted";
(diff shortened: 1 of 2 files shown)


YouveGotMeowxy:
Can anyone help me with this one? I'm using the current glances subdomain conf and still get a 502 Bad Gateway error. My Compose looks like this, but I cant find what could be the issue?

```
  glances:
    hostname: 'glances'
    image: 'nicolargo/glances:alpine-latest-full'
    pid: 'host'
    environment:
      GLANCES_OPT: '-w'
    volumes:
      - '/opt/docker/configs/glances/glances.conf:/glances/conf/glances.conf:ro'
      - '/var/run/docker.sock:/var/run/docker.sock:ro'
    network_mode: 'host'
    ports:
      - '127.0.0.1:61208:61208'
      - '127.0.0.1:61209:61209'
    deploy:
      mode: 'replicated'
      replicas: 1
      placement:
        constraints:
          - 'node.labels.Main == true'
      restart_policy:
        condition: 'any'
```

I suspect it's something obvious that I'm just missing?

Also just a side question; do I need the "full" version of glances in this case, or will it run fine with the "small" version?

TIA!

YouveGotMeowxy:
> Don't run it in host networking

ok, thank you, I will try that. Will that remove the stats that glances is supposed to show for the host machine?

aptalca:
I don't know as I don't use it. But if you do use host networking, then you need to set `upstream_app` in the proxy conf to the host's lan ip instead of the container name

PR #5, opened 2018-12-04
Add ipv6 support

nemchik:
Follow up from https://github.com/linuxserver/docker-letsencrypt/pull/180


aptalca:
Looks good, but has it been tested extensively? I have no ipv6 experience or testing capability

PR #6, opened 2018-12-05
Add frame header hide for pihole

nemchik:
Same frame thing that portainer has going on.


aptalca:
Out of curiosity, what was the original reason for separating the root location from the /admin location?

On my server I only have the root location defined for proxy and it works just fine.

nemchik:
You can add auth to the admin block and skip it on the root location.

aptalca:
ah, gotcha

PR #10, opened 2018-12-27
added config for grocy

homerr:

### DIFF
--- a/grocy.subdomain.conf.sample
+++ b/grocy.subdomain.conf.sample
@@ -1,0 +1,31 @@
+# make sure that your dns has a cname set for grocy
+
+server {
+    listen 443 ssl;
+    listen [::]:443 ssl;
+
+    server_name grocy.*;
+
+    include /config/nginx/ssl.conf;
+
+    client_max_body_size 0;
+
+    # enable for ldap auth, fill in ldap details in ldap.conf
+    #include /config/nginx/ldap.conf;
+
+    location / {
+        # enable the next two lines for http auth
+        #auth_basic "Restricted";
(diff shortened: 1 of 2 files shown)


grocy.subdomain.conf.sample
```diff
@@ -0,0 +1,31 @@
...
+        #auth_basic_user_file /config/nginx/.htpasswd;
+
+        # enable the next two lines for ldap auth
+        #auth_request /auth;
+        #error_page 401 =200 /login;
+
+        include /config/nginx/proxy.conf;
+        resolver 127.0.0.11 valid=30s;
+        set $upstream_grocy grocy;
+        proxy_pass http://$upstream_grocy:80;
```
MrTarantula: The LinuxServer readme specifies 9283 as a recommended port (although grocy runs on 80 internally).

j0nnymoe: Because when running this reverse proxy and on a bridge, it'll use port 80.

Akruidenberg:
How to get this working? Linuxserver Grocy has allreally an NGINX server inside. Ik would like to use the letsencrypt container for this.

grocy.subdomain.conf.sample
Akruidenberg: Fixed with 9283 as port.

aptalca:
@homerr I cannot get the subfolder method working no matter what I try. The subdomain works with no problems (without a base url)

aptalca:
@homerr can you remove the subfolder one from this PR and we'll merge the subdomain only? We can then try and figure out how to make subfolder work

CHBMB:
I'm closing this as subfolder isn't working

PR #18, opened 2019-02-08
fix websocket proxying

ryester27:
This should fix the websocket HTTP 301 errors that were present in the dashboard.


aptalca:
Can you elaborate? I don't experience any websocket errors with the current config.

aptalca:
ahh, nevermind, I see the errors with the subfolder method, but not the subdomain, but it's prudent to apply this to both.

Thanks for the PR

PR #19, opened 2019-02-18
Change upstream variable for new unifi-controller name.

CHBMB:
As we're deprecating the old image and the new container is called unifi-controller.

@aptalca Unless you want to tackle this in a different way?


aptalca:
the container name for the new one is `unifi-controller`. Perhaps we should leave this one as is and create a new one with the name `unifi-controller.subdomain.conf`? Or we can just rename this as well. I'm fine either way

PR #21, opened 2019-03-13
Add proxy conf example for mytinytodo

breakall:

### DIFF
--- a/mytinytodo.subfolder.conf.example
+++ b/mytinytodo.subfolder.conf.example
@@ -1,0 +1,11 @@
+# set the mtt_url config in db/config.php
+
+location ^~ /todo {
+    return 301 $scheme://$host/todo/;
+}
+
+location ^~ /todo/ {
+    include /config/nginx/proxy.conf;
+    set $upstream_mytinytodo mytinytodo;
+    proxy_pass http://$upstream_mytinytodo:80/;
+}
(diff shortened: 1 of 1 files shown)


mytinytodo.subfolder.conf.example
```diff
@@ -0,0 +1,12 @@
+# works with https://github.com/breakall/mytinytodo-docker
+# set the mtt_url config in db/config.php
```
aptalca: Perhaps should read `# set the mtt_url config to todo in db/config.php`?

breakall:
You need to enter the full URL that serves as the base for the mytinytodo installation into the value for that config setting. 

e.g. ‘https://sub.domain.com/todo/‘

The key is that the user should go into that config file and fill out that setting according to the instructions.

breakall:
> Please add the optional auth bits for ldap and http auth like here:
> https://github.com/linuxserver/reverse-proxy-confs/blob/master/airsonic.subfolder.conf.sample#L4-L10
> 
> And make sure to add the resolver as well: https://github.com/linuxserver/reverse-proxy-confs/blob/master/airsonic.subfolder.conf.sample#L13

Will do.

breakall:
I'm new to contributing to other projects on github. I've changed the code in my fork addressing your review notes-- please let me know if I need to do anything else.

aptalca:
No worries, you're doing it the right way with the edits.

Please move the auth lines into the second location block like the example in the link.

With regards to the info at the top, as a general rule, be as specific as possible. For instance, the conf expects the sub folder (url base) in mytinytodo to be set to /todo and if it's set to something else it won't work. You need to specify that otherwise people will set theirs to mytinytodo or whatever they feel like and then come here to say it's not working.

PR #22, opened 2019-03-14
Add resiliosync subdomain proxy-conf for nginx

mcaron1234:


nemchik:
Are hyphens valid in nginx variable names such as `$upstream_resilio-sync` ?

mcaron1234:
Not sure about the hyphen in the servername, I don't love resilio-sync.example.com vs resiliosync.example.com, but this way does match the container name.  

I left out the hyphen in the variable name, but added it to the variable value.  My googling suggests it may be possible to have a variable name in nginx with a hyphen but it requires a bit of coding that was over my level.

nemchik:
The `server_name` should be fine with the hyphen, as should the value of the variable. If nginx variables are anything like bash they do not support hyphens in the variable name (because hyphen is a character used for other things).

If that is the case (if nginx does not allow hyphens in variable names) then the variable name will be inconsistent with the container name (value of the variable). My OCD really wants everything to match but it's not my call.

The only other config where this does natch match is nzbhydra, which could easily be renamed to hydra2. Aside from that the unifi-controller config also used a hyphen (just a side note in case this discussion leads to some kind of decision about hyphens).

aptalca:
Hyphen is ok in server name and variable value, but not variable name.

mcaron1234:
Great. This should be all set now, then.

aptalca:
Looks good, but to satisfy my OCD, can you add the hyphen to the filename as well? Then I'll merge

aptalca:
I made the change, this one's ready to go

mcaron1234:
Thanks. I’ve been away for work for a few days.

aptalca:
no worries, I didn't want to make the change myself so I could approve and merge, but now I gotta wait for another team member :-)

jkossis:
Is there any reason for there not to be a corresponding subfolder proxy conf as well?

aptalca:
The reason is either no one bothered to create one, or they tried and couldn't get it to work. Subfolder proxy is very tricky if the upstream app doesn't support a base url

PR #24, opened 2019-03-19
Added proxy_pass_header  X-Transmission-Session-Id;

daan1024:
I added "proxy_pass_header  X-Transmission-Session-Id;" to the tranmission subfolder config, it might also be needed for the subdomain, I simply don't know.

I had to add this to my config because I was getting a 409: Conflict error with the message that I had to pass this header to fix the error, I did and it fixed it.
I use rpc authentication, this might be the reason that I need this, but if I need to others might aswell


daan1024:
I've made the requested changes

PR #26, opened 2019-03-21
added missing proxy.conf and resolver for airsonic subdomain

alex-phillips:

### DIFF
--- a/airsonic.subdomain.conf.sample
+++ b/airsonic.subdomain.conf.sample
@@ -14,8 +14,6 @@ server {
     #include /config/nginx/ldap.conf;
 
     location / {
-        include /config/nginx/proxy.conf;
-
         # enable the next two lines for http auth
         #auth_basic "Restricted";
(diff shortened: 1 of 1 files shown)


airsonic.subdomain.conf.sample
```diff
@@ -14,6 +14,9 @@ server {
     #include /config/nginx/ldap.conf;
 
     location / {
+        include /config/nginx/proxy.conf;
+        resolver 127.0.0.11 valid=30s;
```
CHBMB: This is a duplicate of line 25/28

aptalca:
To satisfy my OCD, can you move the proxy.conf line to right above the resolver line so it's in line with the rest of the confs?

CHBMB:
> To satisfy my OCD, can you move the proxy.conf line to right above the resolver line so it's in line with the rest of the confs?

To piss @aptalca off and satisy my child like sense of mischievousness.  Can we leave it where it is? :grinning:

PR #31, opened 2019-04-15
Add Kanzi configs

CHBMB:
aptalca is a legend, a true god amongst men.  All bow to aptalca...... :D


aptalca:
with that comment, I have to merge it :-)

PR #34, opened 2019-04-23
Enable HTTP Strict Transport Security.

christiananguiano:
This is added for security purposes and in order to suppress the "The "Strict-Transport-Security" HTTP header is not set to at least "15552000" seconds." warning in Nextcloud's security scan.

### DIFF
--- a/nextcloud.subdomain.conf.sample
+++ b/nextcloud.subdomain.conf.sample
@@ -21,7 +21,9 @@ server {
     include /config/nginx/ssl.conf;
 
     client_max_body_size 0;
-
+    
+    add_header Strict-Transport-Security "max-age=15552000; includeSubDomains";
+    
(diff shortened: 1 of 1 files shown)


nemchik:
This is done in ssl.conf which is included in this file. No need to do it here. Just uncomment the line in ssl.conf.

aptalca:
https://media.giphy.com/media/js3SsYYvMiWLC/giphy.gif

HSTS with subdomains is a sledgehammer that should only be enabled by the users themselves: https://github.com/linuxserver/docker-letsencrypt/blob/master/root/defaults/ssl.conf#L20

aptalca:
@christiananguiano thanks for the PR, though

christiananguiano:
Thanks for that info!

PR #37, opened 2019-05-18
ttrss subdomain

CHBMB:
I'm not sure if subfolder is actually possible, and I don't use ttrss myself, but this may be helpful to some.

### DIFF
--- a/ttrss.subdomain.conf.sample
+++ b/ttrss.subdomain.conf.sample
@@ -1,0 +1,41 @@
+# make sure that your dns has a cname set for ttrss and your container name is ttrss
+# In tt-rss config.php the SELF_URL_PATH should be set to http://ttrss.server.com/ NOT https://ttrss.server.com although it will be accessible on https://ttrss.server.com
+
+server {
+    listen 80;
+        server_name ttrss.*;
+        return 301 https://$server_name$request_uri;
+}
+
+server {
+    listen 443 ssl;
+    listen [::]:443 ssl;
+
+    server_name ttrss.*;
+
+    include /config/nginx/ssl.conf;
+
+    client_max_body_size 0;
+
+    # enable for ldap auth, fill in ldap details in ldap.conf
+    #include /config/nginx/ldap.conf;
+
+    location / {
+        # enable the next two lines for http auth
+        #auth_basic "Restricted";
+        #auth_basic_user_file /config/nginx/.htpasswd;
+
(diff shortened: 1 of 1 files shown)


nemchik:
Does the hyphen conversation from https://github.com/linuxserver/reverse-proxy-confs/pull/22 apply here?

aptalca:
not a hyphen issue, but it seems we used the wrong container name. I'll PR a fix

ryester27:
Serving from a subfolder should be possible, but I haven't had much luck passing the sanity checks with my own attempts at writing a nginx config. I'll give it another try now that there's a confirmed working config that I can look at

nemchik:
I've got a mostly working subfolder conf, but when you login it redirects you to the domain without the subfolder. The login does work, you just have to go back to the subfolder URL manually.

ryester27:
> I've got a mostly working subfolder conf, but when you login it redirects you to the domain without the subfolder. The login does work, you just have to go back to the subfolder URL manually.

That's about been my experience as well. There's also some other oddities like the "close this window" button not working on some windows, and unsubscribing from a feed throws an exemption.

I'd like to see if I can get this working without any config changes to the linuxserver/tt-rss container, but it's very much possible that tt-rss does need to run from within a subfolder at the tt-rss container's webserver root

nemchik:
The config.php explicitly lists `http://example.org/tt-rss/` for `SELF_URL_PATH` so I'm assuming they at least have accounted for the idea of subfolder proxies in some way, but I still get the odd behavior. Here's what I've come up with https://github.com/nemchik/reverse-proxy-confs/blob/ce084952ea6431c66dc8c239066cc9a44234229d/tt-rss.subfolder.conf.sample

ryester27:
> The config.php explicitly lists `http://example.org/tt-rss/` for `SELF_URL_PATH` so I'm assuming they at least have accounted for the idea of subfolder proxies in some way, but I still get the odd behavior. Here's what I've come up with https://github.com/nemchik/reverse-proxy-confs/blob/ce084952ea6431c66dc8c239066cc9a44234229d/tt-rss.subfolder.conf.sample

The fact that's there's a sanity check at all for the SELF_URL_PATH value leads me to believe that if you want tt-rss to be served from a subfolder, that it must reside in a subfolder at the document root. Rewriting URI requests only goes so far.

Anyways, we should probably create a separate issue to continue discussion.

PR #42, opened 2019-05-26
Create gitea.subfolder.conf.sample

anoma:
Thanks to chbmb for the fix. Based on existing work from this repo. Copyright passed to the project to with as they wish.


CHBMB:
Can confirm as working.

PR #47, opened 2019-06-01
Add ZNC config

bamtan:
This adds support for Linuxserverio's ZNC container (https://forums.unraid.net/topic/42908-support-linuxserverio-znc/).

I haven't created a subfolder.sample as I'm unable to test and verify.

### DIFF
--- a/znc.subdomain.conf.sample
+++ b/znc.subdomain.conf.sample
@@ -1,0 +1,31 @@
+# make sure that your dns has a cname set for znc
+
+server {
+    listen 443 ssl;
+    listen [::]:443 ssl;
+
+    server_name znc.*;
+
+    include /config/nginx/ssl.conf;
+
+    client_max_body_size 0;
+
+    # enable for ldap auth, fill in ldap details in ldap.conf
+    #include /config/nginx/ldap.conf;
+
+    location / {
+        # enable the next two lines for http auth
+        #auth_basic "Restricted";
+        #auth_basic_user_file /config/nginx/.htpasswd;
+
+        # enable the next two lines for ldap auth
+        #auth_request /auth;
+        #error_page 401 =200 /login;
+
+        include /config/nginx/proxy.conf;
+        resolver 127.0.0.11 valid=30s;
+        set $upstream_znc znc;
+        proxy_pass https://$upstream_znc:6501;
+		proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+    }
+}
--- a/znc.subdomain.conf.sample
+++ b/znc.subdomain.conf.sample
@@ -26,6 +26,5 @@ server {
         resolver 127.0.0.11 valid=30s;
         set $upstream_znc znc;
         proxy_pass https://$upstream_znc:6501;
-		proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
     }
 }


CHBMB:
@bamtan Thanks, we'll get this tested our end and hopefully merge.

znc.subdomain.conf.sample
```diff
@@ -0,0 +1,30 @@
...
+        #auth_basic_user_file /config/nginx/.htpasswd;
+
+        # enable the next two lines for ldap auth
+        #auth_request /auth;
+        #error_page 401 =200 /login;
+
+        include /config/nginx/proxy.conf;
+        resolver 127.0.0.11 valid=30s;
+        set $upstream_znc znc;
+        proxy_pass https://$upstream_znc:6501;
```
anoma: Ideally this `proxy_pass` should be to HTTP and not HTTPS since by default it will be self signed. I realise we are inconsistent in this approach but we might as well start somewhere.

CHBMB: Will that require a change our end in [docker-znc](https://github.com/linuxserver/docker-znc) though?

anoma: I dont know. Certainly not based on my setup here but it is entirely possible I made an enhancement  to the default config there too. I will check asap. Also I have subfolder working so we can push this one right through to completion.

CHBMB: By default it's over SSL

https://github.com/linuxserver/docker-znc/blob/master/root/defaults/znc.conf#L16

anoma:
You know my opinion on shipping self signed SSL. On my setup here I have a a seperate listener for IRC and the GUI and using
```
        AllowIRC = false
        AllowWeb = true
        SSL = false
```
I deploy a separate non SSL we GUI for management. I also have real SSL cross volume mounted from the LetsEncrypt container for the IRC port. 

Lets take this conversation offline because its part of a larger discussion

znc.subdomain.conf.sample
anoma: You know my opinion on shipping self signed SSL. On my setup here I have a a seperate listener for IRC and the GUI and using
```
        AllowIRC = false
        AllowWeb = true
        SSL = false
```
I deploy a separate non SSL we GUI for management. I also have real SSL cross volume mounted from the LetsEncrypt container for the IRC port. 

Lets take this conversation offline because its part of a larger discussion

nemchik: A quick search of this repository brought me results for 11 other proxy configs that use https in the `proxy_pass` line. This list includes Heimdall, Kanzi, Nextcloud, Unifi-Controller, and more. I don't know exactly how https is being handled by each one (self signed, or some other way), but it seems like this isn't an uncommon practice. I see no issue with this proxy config as is assuming that it works with the LSIO znc image the way all other proxies included in this repo that have a corresponding LSIO image work out of the box. (p.s. I have not tested it, just basing my opinion on the syntax matching the rest in the repo).

CHBMB: @bamtan

Can I request, in line with best practice, that you configure it with http and add comments to the top of the file on how to configure ZNC to run as such.

Then I'll approve and merge.

Thanks for the PR.

@xe Want to do the same for subfolder?

bamtan: @CHBMB 

With the config file set to http on line 28, I get `502 Bad Gateway` for ZNC. I don't have the knowledge or understanding to get ZNC to run with the config set to http. 

Is it more to do with ZNC's configuration side of things? Because I haven't touched anything in that regard, either. 

Thanks for the review.

Xe: I'm sorry but how am I relevant?

anoma: Wrong XE tagged. Feel free to ignore.

CHBMB: @xe Sorry mate, my mistake.  Apologies.

CHBMB: @bamtan No worries I'll sort it out.

PR #48, opened 2019-06-03
Fix comments on Plex configs

nemchik:
The instruction comments say to replace the line with `https`, but later in the config file it uses `http` (which is correct).


nemchik:
I caught a second error while re-reading my PR changes (sanity check) and added that to this as well. `https://yourdomain.url/plex:443` is not valid.

PR #49, opened 2019-06-04
Fix mytinytodo formatting

nemchik:
The top location block is adjusted to match the formatting of other location blocks performing the same redirect across the repo. The commented lines are adjusted to remove a space from the lines that have things that could be uncommented and used (where actual comments retain the space).

I came across this by running:
```bash
find . -type f -print0 | xargs -0 grep "\sauth_request /auth"
find . -type f -print0 | xargs -0 grep "\sauth_basic \"Restricted\""
```
Where I was actually trying to find configs in which I had enabled auth, but this config returned in the results due to the different formatting.


CHBMB:
@nemchik I don't understand the logic removing the `^~`

PR #50, opened 2019-06-09
smokeping configs

lloydpick:
Smokeping configs for subdomain or subfolder. Configured as per the default of the linuxserver/smokeping container


aptalca:
@lloydpick thanks for the PR

PR #55, opened 2019-06-10
Remove ^~

nemchik:
http://nginx.org/en/docs/http/ngx_http_core_module.html#location

> If the longest matching prefix location has the “^~” modifier then regular expressions are not checked.

I've run into a issue specifically with the nzbget proxy config where this `^~` modifier prevented nzb360 from accessing the RPC URLs when an `auth_request` directive was used in the main location block (ex: ldap or Organizr auth). I have tested about 1/4 of these, but based on the documentation I don't see a need for them. It seems like these were mostly included because a small number of configs had them (traced back to radarr https://github.com/linuxserver/docker-letsencrypt/commit/ac896d056e3fbd6b5e25d7506a32bcd04dd7db65, which I have confirmed doesn't need it) and many configs created after that just did a copy/replace job (I did many of them) and stayed in line with existing formatting in the repo.

If this isn't going to be accepted I'll make a PR specifically for only nzbget.

### DIFF
--- a/airsonic.subfolder.conf.sample
+++ b/airsonic.subfolder.conf.sample
@@ -1,6 +1,6 @@
 # set the CONTEXT_PATH variable to /airsonic in airsonic container.
 
-location ^~ /airsonic {
+location /airsonic {
     # enable the next two lines for http auth
     #auth_basic "Restricted";
     #auth_basic_user_file /config/nginx/.htpasswd;
--- a/bazarr.subfolder.conf.sample
+++ b/bazarr.subfolder.conf.sample
@@ -3,7 +3,7 @@
 location /bazarr {
     return 301 $scheme://$host/bazarr/;
 }
-location ^~ /bazarr/ {
+location /bazarr/ {
     # enable the next two lines for http auth
     #auth_basic "Restricted";
     #auth_basic_user_file /config/nginx/.htpasswd;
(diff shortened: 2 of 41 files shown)


aptalca:
The `^~` modifier is needed especially for subfolder configs. One common scenario is, a php request of a proxied php based app will incorrectly passed to letsencrypt container's php parser rather than the request sent to the proxied app because php regex will take priority over a basic subfolder location in the same server block. Many other scenarios are possible in more complex configs. The `^~` modifier makes the location a higher priority than a regex match, properly proxying all requests at that subfolder.

nemchik:
I see. I comment out the PHP block in the default site conf to prevent what you described. I know that solution isn't for everyone.

I tried a bunch of different config combinations with nzbget such as not using regex for the RPC locations (which meant having 3) but removing the `^~` was what got it working.

I'm open to other ideas to resolve the issue with nzbget.

nemchik:
Ok, I've got a config for nzbget working with the `^~` using 3 separate locations. I can submit that and scrap this.

This does bring up the question "what about the configs that lack `^~` currently?" Should they be adjusted to include it? Just wondering about consistency.

aptalca:
If it ain't broke. . .

nemchik:
Fair enough. I'll close this and prepare a separate PR for only the nzbget fix.

PR #57, opened 2019-06-10
Jellyfin

CHBMB:
Subdomain and subfolder.


nemchik:
Maybe affected by #55

CHBMB:
@nemchik I'll wait for @aptalca to comment on this, but my gut feeling on it, is if the config starts with: 
```

location /application {
    return 301 $scheme://$host/application/;
}
```
then I *think* it's required.  But I'm perfectly happy to be proved wrong because I'm just not sure.

nemchik:
I have confirmed that is not the case. The Plex proxy redirects just fine without `^~` (cleared browser cache and checked with incognito).

CHBMB:
Fair enough.  But I'm 100% convinced I don't know enough to approve/merge this so I'm going to defer to yourself and @aptalca :wink:

nemchik:
Agreed.

PR #59, opened 2019-06-10
Adjust locations for nzbget rpc

nemchik:


nemchik:
Follow up to conversations from #55

PR #64, opened 2019-06-18
Add pyload subdomain

nemchik:
I can't get subfolder working (and according to a number of issues on their git repo there's plenty of others having the same trouble).

Subdomain works fine though, tested and confirmed.


aptalca:
I think you PR'ed the wrong file, it's all for bazarr ;-)

nemchik:
Fixed.

PR #76, opened 2019-07-22
Add non-auth locations for qbittorrent

nemchik:
These locations allow nzb360 to access qbit without auth (same as most other `/api` locations in other confs).

### DIFF
--- a/qbittorrent.subdomain.conf.sample
+++ b/qbittorrent.subdomain.conf.sample
@@ -30,4 +30,59 @@ server {
         proxy_set_header Referer '';
         proxy_set_header Host $upstream_qbittorrent:8080;
     }
+
+    location ^~ /qbittorrent/api {
+        include /config/nginx/proxy.conf;
+        resolver 127.0.0.11 valid=30s;
+        set $upstream_qbittorrent qbittorrent;
+        rewrite /qbittorrent(.*) $1 break;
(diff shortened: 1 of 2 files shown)


anoma:
@nemchik can you point me at anything that explains how access control works for the api now?

I dont see anything about api keys but find it hard to believe this change allows just anyone `/api` access.

nemchik:
@anoma https://github.com/qbittorrent/qBittorrent/wiki/Web-API-Documentation#authentication does provide some information about authorizing to the `/api` endpoints. In my testing with nzb360 the "test connection" button on the torrent settings screen in the app will always complete successfully even if you use incorrect credentials. You will eventually get ip banned for testing with incorrect credentials too many times, and then the test connection button will fail. This seems to be a bug with nzb360 that I will report to the dev.

Anyway, if you set incorrect credentials and then return to the actual app to manage torrents it will fail asking you to retry (which counts against the number of failed logins mentioned above). If you set the correct credentials then it will show the torrent list accordingly.

It seems that the current version of qbit has moved everything under `/api` however their wiki documents previous versions of qbit's API here https://github.com/qbittorrent/qBittorrent/wiki/WebUI-API-Documentation and here https://github.com/qbittorrent/qBittorrent/wiki/WebUI-API-Documentation-(qBittorrent-v3.1.x) where you can see the other locations specified in this PR are used.

PR #88, opened 2019-10-03
Fix tt-rss subdomain redirection

Dreytac:
I was having an issue where attempting to login to tt-rss was redirecting me to "https://tt-rss.%2a/index.php" instead of the expected "https://tt-rss.mydomain.com/index.php". I tracked the issue down to this line in the tt-rss.subdomain.conf file.

I believe $host and not $server_name should be used here.


aptalca:
Actually, that port 80 redirect is not supposed to be there at all. We do the redirect in the default site conf. I don't know how that got in there.

Dreytac:
That makes sense. I've updated the PR to remove the port 80 redirect entirely.

PR #89, opened 2019-10-07
Add Nextcloud recommended security options

Dreytac:
There were 2 security options recommended by Nextcloud under the "Security & setup warnings" page. I have implemented both of those recommendations into this config.

### DIFF
--- a/nextcloud.subdomain.conf.sample
+++ b/nextcloud.subdomain.conf.sample
@@ -21,6 +21,9 @@ server {
     include /config/nginx/ssl.conf;
 
     client_max_body_size 0;
+    
+    add_header X-Frame-Options "SAMEORIGIN";
+    add_header Strict-Transport-Security "max-age=15552000" always;
 
     location / {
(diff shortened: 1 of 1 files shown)


Dreytac:
Does enabling them in ssl.conf have any adverse affects on the other proxy configs? I didn't enable them there due to worries about it affecting other configurations (and that it was explicitly recommended by Nextcloud).

aptalca:
HSTS has various potential adverse effects. Especially when it's used with "enable subdomains" and "preload". Browsers cache HSTS very aggressively, even a cache clear is not enough to get rid of it.

We like to make it totally optional so that only if the user is sure about enabling it, they can. But it has to be done by the user, not by default.

I don't think our nextcloud image benefits from it much, if at all, since it's not even listening on port 80.

Dreytac:
That's what I thought. I'll keep the options local to my config just to satisfy my desire to have a clear security page if nothing else.

PR #98, opened 2019-11-12
Remove resolver from proxy confs

nemchik:
Follow up to https://github.com/linuxserver/docker-letsencrypt/pull/317

Since `resolver 127.0.0.11 valid=30s;` was added to `ssl.conf` it was brought to my attention that it is redundant to be in all of the proxy configs.

There's also a few whitespace and EOL changes in here (blame vscode, I got tired of trying to stop it).

### DIFF
--- a/adguard.subdomain.conf.sample
+++ b/adguard.subdomain.conf.sample
@@ -23,7 +23,6 @@ server {
         #error_page 401 =200 /login;
 
         include /config/nginx/proxy.conf;
-        resolver 127.0.0.11 valid=30s;
         set $upstream_adguard adguard;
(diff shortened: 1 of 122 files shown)


aptalca:
Right, but since ssl.conf is user configurable, if and when the users change the resolver to 8.8.8.8 or something else, it would break all proxy confs without the user's realization. It would be quite hard for us to troubleshoot that as we don't commonly ask for any modifications to ssl.conf

nemchik:
I would argue that all the files in the config volume are "user configurable", but i can agree it's more likely that someone would change ssl.conf and break their stuff and not worth the support queries.

PR #106, opened 2019-12-17
Fix Airsonic subdomain webUI http request on 443

### LINKED ISSUE #105: Airsonic creating http request on 443
Issue:  
Airsonic streaming when using a subdomain redirection issue.  

When streaming via the webui on https://airsonic.myaddress.com opening a stream generates a url that looks like http://airsonic.myaddress.com:443/stream?player=5&id=176
Which nginx sees as a bad request as it's a http request on port 443.

Davis-A:
Issue:  Fixes #105 
When streaming via the webui on https://airsonic.myaddress.com opening a stream generates a url that looks like http://airsonic.myaddress.com:443/stream?player=5&id=176
Which nginx sees as a bad request as it's a http request on port 443.  

Adding `proxy_set_header X-Forwarded-Host  $http_host;` fixes this issue.

[Note the Airsonic docs list X-Forwarded-Host as a required parameter when running it behind a reverse proxy](https://airsonic.github.io/docs/proxy/prerequisites/)

### DIFF
--- a/airsonic.subdomain.conf.sample
+++ b/airsonic.subdomain.conf.sample
@@ -25,6 +25,7 @@ server {
         include /config/nginx/proxy.conf;
         resolver 127.0.0.11 valid=30s;
         set $upstream_airsonic airsonic;
+        proxy_set_header X-Forwarded-Host  $http_host;
         proxy_pass http://$upstream_airsonic:4040;
     }
 }


nemchik:
We've got https://github.com/linuxserver/docker-letsencrypt/blob/c943d13c535cf3ff118e52489d0bc6efb2655eb4/root/defaults/proxy.conf#L22 which is included but slightly different.

Here a quick summary of the difference with `$host` and `$http_host` https://stackoverflow.com/questions/15414810/whats-the-difference-of-host-and-http-host-in-nginx and another good link for info https://serverfault.com/questions/706438/what-is-the-difference-between-nginx-variables-host-http-host-and-server-na

Is there a specific reason you're waiting the `$http_host` variable?

Davis-A:
>Is there a specific reason you're waiting the $http_host variable?

No reason in particular.  I'm new to setting this stuff up.

I started with the `airsonic.subdomain.conf.sample` from this repo and hit the issue mentioned.  

Reading the Airsonic docs the only proxy\_set\_header not included in the sample was X-Forwarded-Host and the [Airsonic docs for nginx](https://airsonic.github.io/docs/proxy/nginx/) example is `proxy_set_header X-Forwarded-Host  $http_host;` which resolved the issue.

nemchik:
Try deleting your proxy.conf and removing the `proxy_set_header` you added to the airsonic config and restart the container. It should pull a new copy of the proxy.conf for you. If it works then we don't need to modify the existing airsonic proxy :smile:

aptalca:
@Davis-A sounds like you had your letsencrypt container for a while. Those user customizable conf files under config do not get updated when you pull a newer image and recreate the container because we don't want to overwrite and revert user changes. That's why they now contain a link in their first line, which shows you all the changes made to them by date so you can manually apply the same modifications.

Alternatively, you can delete them and restart the container and newest default versions will be created.

aptalca:
Closing this now. You can reopen if there are still issues

PR #109, opened 2020-01-01
removed redundant resolver line

drizuid:
removed redundant resolver 127.0.0.11 valid=30s; from server block. It's already in ssl.conf which is included via line 9


drizuid:
potentially need to add the resolver config under each location subsection to account for user-made changes in ssl.conf to the resolver there.

PR #117, opened 2020-02-04
reverse proxy config for shaarli

JOduMonT:
reverse proxy config for shaarli


aptalca:
What about http auth and ldap auth?

j0nnymoe:
Still missing http auth?

JOduMonT:
yes I prefer to not contribute and keep it for myself

PR #119, opened 2020-02-13
Create statping.subdomain.conf.sample

The-Kid-Gid:
Create subdomain conf for [https://github.com/hunterlong/statping](url).

### DIFF
--- a/statping.subdomain.conf.sample
+++ b/statping.subdomain.conf.sample
@@ -1,0 +1,31 @@
+# make sure that your dns has a cname set for statping and that your statping container is not using a base url
+# the default url is status.yourdomain.tld, but you can change it to whatever you want
+
+server {
+    listen 443 ssl;
+    listen [::]:443 ssl;
+
+    server_name status.*;
+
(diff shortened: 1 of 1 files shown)


statping.subdomain.conf.sample
```diff
@@ -0,0 +1,31 @@
+# make sure that your dns has a cname set for statping and that your statping container is not using a base url
+# the default url is status.yourdomain.tld, but you can change it to whatever you want
+
+server {
+    listen 443 ssl;
+    listen [::]:443 ssl;
+
+    server_name status.*;
```
aptalca: Let's set this to `statping.*` to standardize with the rest of the proxy confs

aptalca:
Thanks for the PR. Can you also add a blurb about the container name set to `statup`. It's not very clear on the statping docs (only the compose stack references it, but not the other samples).

statping.subdomain.conf.sample
The-Kid-Gid: Done

PR #120, opened 2020-02-14
NZBGet URL Credentials

HalianElf:
This allows the NZBGet credentials to be passed through the URL since LunaSea changed to using this method for auth instead of headers.

### DIFF
--- a/nzbget.subdomain.conf.sample
+++ b/nzbget.subdomain.conf.sample
@@ -28,21 +28,21 @@ server {
         proxy_pass http://$upstream_nzbget:6789;
     }
 
-    location ~ ^/nzbget(/.*)?/jsonrpc {
+    location ~ (/nzbget)?(/[^\/:]*:[^\/:]*)?/jsonrpc {
         include /config/nginx/proxy.conf;
         resolver 127.0.0.11 valid=30s;
         set $upstream_nzbget nzbget;
         proxy_pass http://$upstream_nzbget:6789;
     }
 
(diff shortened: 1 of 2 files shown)


nemchik:
What would be an example of a URL that would match these?

HalianElf:
It would match `/nzbget/username:password/jsonrpc` which is how LunaSea is doing it now.

nemchik:
I knew I had heard of this before :smile:
https://github.com/nzbget/nzbget/issues/577#issuecomment-435603360

That being said I'd like to do something more similar to what we have for sonarr (and others) here:
https://github.com/linuxserver/reverse-proxy-confs/blob/a91a99b4b58b220a5f3c2e53fe819bc00ecf49d8/sonarr.subdomain.conf.sample#L31

Specifically:
`location ~ (/nzbget)?(/[^\/:]*:[^\/:]*)/jsonrpc {`

Would you modify your PR to this pattern and include both subdomain and subfolder?

HalianElf:
Changed it to your suggested regex (plus the `?` you were missing)

nemchik:
Good catch on the missing `?`

For the password we could do
`location ~ (/nzbget)?(/[^\/:]*:[^\/]*)?/jsonrpc {`
Essentially we know the username and password are split by a `:` and I assume can't have slashes in them.

HalianElf:
Yea, I did test that too. `/` do break it (they work for the header but not in the url)

PR #124, opened 2020-02-16
added osticket.subdomain.conf

zellkou:

### DIFF
--- a/osticket.subdomain.conf.sample
+++ b/osticket.subdomain.conf.sample
@@ -4,7 +4,7 @@ server {
     listen 443 ssl;
     listen [::]:443 ssl;
 
-    server_name help.*;
+    server_name osticket.*;
 
     include /config/nginx/ssl.conf;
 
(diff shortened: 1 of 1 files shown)


aptalca:
Thanks for the PR, but why would the port number be changed?

zellkou:
That was a note to myself left in by mistake. I already had the container running before setting up letsencrypt so instead of moving it to the same docker network I just used  $HOSTIP$Port instead of $upstream_osticket:80 in my personal config.

aptalca:
Gotcha, thanks. And one last thing. Please change the subdomain to `osticket` from `help` so it's uniform. Then we'll merge.

zellkou:
No problem, I just went with what was recommend in the osticket documentation for the subdomain name.

aptalca:
I understand. We try to match the subdomain to the app name to prevent any potential collisions between different proxy confs by default. The end users of course can change it to whatever they want when they enable.

Thanks for the changes and the PR.

zellkou:
No problem, I'm just glad I can contribute to the project.

PR #131, opened 2020-02-25
Update organizr.subfolder.conf.sample

19wolf:
location clause was "/" causing errors. Changed to "/organizr"

### DIFF
--- a/organizr.subfolder.conf.sample
+++ b/organizr.subfolder.conf.sample
@@ -1,6 +1,6 @@
 # In order to use this location block you need to edit the default file one folder up and comment out the / location
 
-location / {
+location /organizr {
     # enable the next two lines for http auth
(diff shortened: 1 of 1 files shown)


aptalca:
Actually this is meant to be used for the main homepage `/`, not as subfolder `/organizr`: https://github.com/linuxserver/reverse-proxy-confs/blob/master/organizr.subfolder.conf.sample#L1

19wolf:
When enabled as-is, I get continuous errors on the letsencrypt container and it eventually just fails to proxy anything. What is the fix for that?

nemchik:
The fix is documented in the link aptalca provided https://github.com/linuxserver/reverse-proxy-confs/blob/master/organizr.subfolder.conf.sample#L1 which is line one of the Organizr proxy file you're remaining.

PR #133, opened 2020-02-29
Update organizr.subfolder.conf.sample

19wolf:
Wording might need some tweaking but trying to make it clear there are other options.

### DIFF
--- a/organizr.subfolder.conf.sample
+++ b/organizr.subfolder.conf.sample
@@ -1,4 +1,5 @@
 # In order to use this location block you need to edit the default file one folder up and comment out the / location
+# Alternatively, you can change "location /" to "location/something" in this file
 
 location / {
     # enable the next two lines for http auth


nemchik:
I believe doing this also requires some reconfiguring in the nginx config of the Organizr container.

19wolf:
I don't think so. There's no place within organizr that I saw to "specify subdirectory" like you need to do with ex. sonarr. 

I'm not sure if this file would then need to point to port 9983 [as specified](https://github.com/linuxserver/docker-organizr)? You would definitely need to change the port if you need to change the upstream app section. I have my letsencrypt and organizr in different compose files so they're by default on different networks so I needed to do that, which is specified somewhere?

nemchik:
Our Organizr image is still v1 which the dev no longer updates. They maintain a v2 image. Inside the config is an nginx config that has a `/` location. You would need to adjust that in order to run Organizr on a subfolder.

nemchik:
Closing this. If you need help getting this type of config working please visit our discord or the Organizr discord.

PR #136, opened 2020-03-03
Add SyncLounge subdomain configs

MagicalCodeMonkey:
This adds subdomain configs for SyncLounge.

The SyncLouge container with `latest` or `alpine` tags contains both the client and server applications. Due to this, unless you the `synclounge.subdomain` configuration must have a block for `/slserver`. However, if it is preferred it can be left out and the user will have to use the `synclounge-server.subdomain` configuration in addition if they want to use both.


synclounge.subdomain.conf.sample
```diff
@@ -0,0 +1,71 @@
...
+        include /config/nginx/proxy.conf;
+        resolver 127.0.0.11 valid=30s;
+        set $upstream_app synclounge;
+        set $upstream_port 8088;
+        set $upstream_proto http;
+
+        proxy_set_header Upgrade $http_upgrade;
+        proxy_set_header Connection "upgrade";
+
+        proxy_pass $upstream_proto://$upstream_app:$upstream_port;
```
nemchik: Let's try to keep the formatting
```nginx
        include /config/nginx/proxy.conf;
        resolver 127.0.0.11 valid=30s;
        set $upstream_app synclounge;
        set $upstream_port 8088;
        set $upstream_proto http;

        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection "upgrade";
```

synclounge-server.subdomain.conf.sample
```diff
@@ -0,0 +1,43 @@
+# Use this if you are using the `server` tag for the container, which only contains the SyncLounge server application.
```
nemchik: Maybe we could keep this as just one config if we are good about the documentation in the comments? It could serve both purposes?

synclounge-server.subdomain.conf.sample
```diff
@@ -0,0 +1,43 @@
...
+    server_name synclounge-server.*;
+
+    include /config/nginx/ssl.conf;
+
+    client_max_body_size 0;
+    proxy_redirect off;
+    proxy_buffering off;
+
+    # SyncLounge server defaults the base URL to `/slserver` which doesn't seem to like being set to just `/`
+    location / {
```
nemchik: If the server does not like to run with the base url `/` then how is this supposed to work?

synclounge.subdomain.conf.sample
```diff
@@ -0,0 +1,71 @@
...
+        set $upstream_app synclounge;
+        set $upstream_port 8089;
+        set $upstream_proto http;
+
+        proxy_set_header Upgrade $http_upgrade;
+        proxy_set_header Connection "upgrade";
+
+        proxy_pass $upstream_proto://$upstream_app:$upstream_port/slserver/socket.io/;
+    }
+}
```
nemchik: Missing trailing newline.

synclounge-server.subdomain.conf.sample
MagicalCodeMonkey: It's a bit confusing because SyncLounge server still responds to requests on `/` but it only operates on `/slserver` or whatever the base url (aka `serverroot`) is set to. This still works because it redirects all requests to the server appropriately, users just have to follow the documentation of SyncLounge to make sure the URL is correct when connecting).

With all that said, it would be a moot point if we drop the server subdomain config and just stick with the other one.

synclounge-server.subdomain.conf.sample
MagicalCodeMonkey: Yeah, I was thinking that might be better. Even if the person only uses the server container, it would still work fine

PR #139, opened 2020-03-18
Add Synclounge subfolder conf

MagicalCodeMonkey:
This adds a subfolder config for SyncLounge.

The SyncLouge container with `latest` or `alpine` tags contains both the client and server applications. Due to this, unless you the configuration must have a block for `/slserver`.


MagicalCodeMonkey:
Should be all good now! Thanks!

PR #146, opened 2020-04-10
fix emby websockets

aptalca:


anoma:
I have been running this for a few days and it appears to have fixed all my Emby websocket 404 errors

PR #155, opened 2020-04-25
Add a note about HTTP/2 and Transmission Chrome Extensions

n-i-x:

### DIFF
--- a/transmission.subdomain.conf.sample
+++ b/transmission.subdomain.conf.sample
@@ -1,4 +1,14 @@
-# make sure that your dns has a cname set for transmission
+# Make sure that DNS has a cname set for transmission
+#
+# Some Transmission Chrome extensions cannot handle HTTP/2 proxies as they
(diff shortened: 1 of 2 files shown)


nemchik:
I actually don't think the comment is relevant to the subdomain proxy because it doesn't have `http2` in the `listen` directive.

I believe those were left out intentionally (I'll have to go find the original PR discussion).

nemchik:
https://github.com/linuxserver/docker-letsencrypt/pull/180#issuecomment-441505801

There was the discussion.

Anyway it's not included in the subdomain, but it would affect the subfolder proxy.

n-i-x:
nginx can't support multiple protocols on a single port. Any server block that enables http2 on a port enables http2 on servers listening on the same port. Since the default site in linuxserver/letsencrypt adds http2 to its listen block on 443 it gets enabled across the board to any server block using port 443.

https://github.com/linuxserver/docker-letsencrypt/blob/master/root/defaults/default#L13

PR #163, opened 2020-05-24
add authelia subdomain conf and add auth support

aptalca:


aptalca:
we should merge this before https://github.com/linuxserver/docker-letsencrypt/pull/460 so these get pulled in

PR #174, opened 2020-06-04
include proxy.conf in portainer subdomain config

golles:
I noticed that my general proxy config was not applied to this reverse proxy. All other ones (that I use) do have it included.

I wanted to raise an issue for this, but I assumed it would be easier for everyone to make a PR instead.

I currently have this setup running fine.

### DIFF
--- a/portainer.subdomain.conf.sample
+++ b/portainer.subdomain.conf.sample
@@ -28,6 +28,7 @@ server {
         # enable for Authelia
         #include /config/nginx/authelia-location.conf;
 
+        include /config/nginx/proxy.conf;
         resolver 127.0.0.11 valid=30s;
         set $upstream_app portainer;
         set $upstream_port 9000;
@@ -35,7 +36,6 @@ server {
         proxy_pass $upstream_proto://$upstream_app:$upstream_port;
 
         proxy_set_header Connection "";
-        proxy_http_version 1.1;
         proxy_hide_header X-Frame-Options; # Possibly nott needed after Portainer 1.20.0
     }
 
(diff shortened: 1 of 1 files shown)


portainer.subdomain.conf.sample
```diff
@@ -59,7 +60,6 @@ server {
 
         proxy_set_header Upgrade $http_upgrade;
         proxy_set_header Connection "upgrade";
-        proxy_http_version 1.1;
```
golles, line 62: Without this, nginx will not start, as this is already defined in `/config/nginx/proxy.conf`. I checked for the other fields, they ar not present in `/config/nginx/proxy.conf`

portainer.subdomain.conf.sample
```diff
@@ -28,14 +28,14 @@ server {
         # enable for Authelia
         #include /config/nginx/authelia-location.conf;
 
+        include /config/nginx/proxy.conf;
```
golles, line 31: I added this line before the `resolver` line, like in other conf files

aptalca:
The original conf was pulled directly from portainer docs. They didn't need any of those headers from proxy.conf and we didn't want to risk breaking anything

nemchik:
We do in fact include the proxy.conf in the subfolder config https://github.com/linuxserver/reverse-proxy-confs/blob/207a0ab0f28926c8fc480b2aa8f112f9b15f0d1b/portainer.subfolder.conf.sample#L18  but not in the subdomain config. I'm not against this change as long as it works.

golles:
One thing I forgot to mention is that I have custom settings (for geo IP block) in `proxy.conf` and that was not applied to portainer.
So I think it would be good to update this file, so there is no surprise for anyone else. But, in case you feel like keeping the original provided by portainer team, then I wouldn't mind declining the PR

aptalca:
If you tested and it works, it's all good.

Thanks

PR #183, opened 2020-06-28
Rename youtube-dl.subfolder.conf to youtube-dl.subfolder.conf.sample

CoryEvans2324:
The youtube-dl config file is enabled by default when starting/creating a new docker container.


nemchik:
Good catch. Thanks!

PR #188, opened 2020-07-17
Add code-server subdomain proxy support

freezurbern:
code-server supports proxying ports over a given subdomain.
This commit adds support in docker-letsencrypt.
See https://github.com/cdr/code-server/blob/master/doc/FAQ.md#sub-domains

### DIFF
--- a/code-server.subdomain.conf.sample
+++ b/code-server.subdomain.conf.sample
@@ -39,3 +39,46 @@ server {
         proxy_set_header Connection upgrade;
     }
 }
+
+# Support for code-server subdomain proxy
+# See https://github.com/cdr/code-server/blob/master/doc/FAQ.md#sub-domains
+# This example restricts access to ports 5000 through 5999 of the $upstream_app
+server {
+    listen 443 ssl;
+    listen [::]:443 ssl;
+    
+    server_name ~^(?<subdomain>[5][0-9][0-9][0-9])\.code-server.*;
+
+    include /config/nginx/ssl.conf;
+
+    client_max_body_size 0;
+
+    # enable for ldap auth, fill in ldap details in ldap.conf
+    #include /config/nginx/ldap.conf;
+
+    # enable for Authelia
+    #include /config/nginx/authelia-server.conf;
+
+    location / {
+        # enable the next two lines for http auth
+        #auth_basic "Restricted";
+        #auth_basic_user_file /config/nginx/.htpasswd;
+
+        # enable the next two lines for ldap auth
+        #auth_request /auth;
+        #error_page 401 =200 /ldaplogin;
+
+        # enable for Authelia
+        #include /config/nginx/authelia-location.conf;
+
+        include /config/nginx/proxy.conf;
+        resolver 127.0.0.11 valid=30s;
+        set $upstream_app code-server;
+        set $upstream_proto http;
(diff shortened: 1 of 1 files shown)


code-server.subdomain.conf.sample
```diff
@@ -39,3 +39,46 @@ server {
...
 }
+
+# Support for code-server subdomain proxy
+# See https://github.com/cdr/code-server/blob/master/doc/FAQ.md#sub-domains
+# This example restricts access to ports 5000 through 5999 of the $upstream_app
+server {
+    listen 443 ssl;
+    listen [::]:443 ssl;
+    
+    server_name ~^(?<subdomain>[5][0-9][0-9][0-9])\.code-server.*;
```
aptalca, line 50: Why not just append the main `server_name` with something like `*.code-server.*` so it reverse proxies all sub-subdomains to code-server?

code-server.subdomain.conf.sample
```diff
@@ -39,3 +39,46 @@ server {
         proxy_set_header Connection upgrade;
     }
 }
+
+# Support for code-server subdomain proxy
+# See https://github.com/cdr/code-server/blob/master/doc/FAQ.md#sub-domains
+# This example restricts access to ports 5000 through 5999 of the $upstream_app
```
aptalca, line 45: What is the reason for this limit?

code-server.subdomain.conf.sample
freezurbern: I'm not very familiar with nginx config. I needed this `server_name` to take the port number from the URL and put it into the `proxy_pass` line lower down. For example:
`https://5000.code-server.example.com` -> `http://code-server:5000`

code-server.subdomain.conf.sample
freezurbern: Yes, it is arbitrary. I didn't want to allow all ports to be opened to the destination such as:
`https://443.code-server.example.com` -> `http://code-server:443`
I did this because I had other services running on those ports and didn't want access. Please let me know if there's a better way to do it. This is my first pull request to such a large project.

code-server.subdomain.conf.sample
aptalca: I don't believe that's how that feature works. Code-server is only listening on one port, 8443 in this case. Code-server checks the requested domain and if the last subfolder matches a certain format (ie. port number), code-server itself reverse proxies the app listening on that port and makes it accessible via its port 8443.

In other words, nginx only has to pass on the correct `url`, and code-server does the reverse proxy for the app based on the subdomain.

In this PR, you are effectively trying to recreate what code-server already offers, but via nginx alone.

aptalca: Hmm, I'll try some things locally and will ping back here. Thanks

aptalca: alright, all I had to do was set the `PROXY_DOMAIN` var in code-server docker arguments to `code-server.mydomain.com`, edit the server_name line here to read `server_name code-server.* "~^[0-9]{1,10}\.code-server\..*$";` and then I fired up a simple python webserver listening on port 8000 in code-server console, and the page is served at `https://8000.code-server.mydomain.com`. Of course you also need to make sure that your letsencrypt cert covers the sub-subdomain for code-server otherwise the browser throws errors

PR #191, opened 2020-08-10
homeassistant: enable websockets for entire subd.

aptalca:
necessary for some addons


nemchik:
Saw this and covered it in https://github.com/linuxserver/reverse-proxy-confs/pull/197

PR #201, opened 2020-09-10
Update collabora.subdomain.conf.sample

Zachar2:
It started working for me only after those changes. I also had to turn off Rocket Loader for Collabora subdomain in Cloudflare.

### DIFF
--- a/collabora.subdomain.conf.sample
+++ b/collabora.subdomain.conf.sample
@@ -49,8 +49,22 @@ server {
         set $upstream_proto https;
         proxy_pass $upstream_proto://$upstream_app:$upstream_port;
 
+        proxy_set_header Upgrade $http_upgrade;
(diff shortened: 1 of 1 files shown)


aptalca:
Most if not all of these are already included in SWAG's proxy.conf: https://github.com/linuxserver/docker-swag/blob/master/root/defaults/proxy.conf

Zachar2:
@aptalca You are correct. Weirdest thing is that I removed everything except:

`proxy_set_header Upgrade $http_upgrade;`

which already is in `proxy.conf` and it is in fact working. If I remove that line, it stops functioning again and shows a blank page. I use `linuxserver/nextcloud` and `collabora/code`. Any idea why it has to be repeated here?

Zachar2:
Ok. Major failure here. I had an old version of `proxy.conf` running. Apologies for the fuss.

PR #207, opened 2020-09-22
Update synclounge.subdomain.conf.sample

marissa999:
Added the default path to the proxy_pass


Roxedus:
Why? Nginx should handle this, and it does on my install

aptalca:
This PR would actually break it by telling nginx to proxy `/slserver/slserver`

Issue #198, opened 2020-09-09
Plex websocket handling is incorrect, proposed fix


esticle:
Having recently upgraded from Swag 1.7.0 to 1.8.0 and re-rolling from the sample configs. I noticed that the removal of two lines in [this commit ](https://github.com/linuxserver/reverse-proxy-confs/commit/388a3a7e7fcaa51247311ed9eb106740ba888db1)has broken the websocket connection under 'Alerts' for a server.

A copied Plex sample config has had these two lines removed between v1.7.0 and 1.8.0:

```
proxy_set_header        Upgrade         $http_upgrade;
proxy_set_header        Connection      "upgrade";
```
If you are accessing your Plex instance via Swag/a configuration using these reverse-proxy-confs at version 1.8.0 you will notice that going to Settings -> Alerts you will get the text: **_Unable to listen for events on servername_**.

This is fixed by re-adding the above two lines.

Thanks.

Roxedus:
How outdated is your nginx.conf and proxy.conf?

esticle:
> How outdated is your nginx.conf and proxy.conf?

Thanks for the hint - my nginx.conf was from Swag 1.7.0 which (of course) had [this commit](https://github.com/linuxserver/docker-swag/commit/c67946115a03807ad5399f0a57c9bdbc797d46c7) when it moved to 1.8.0 to sync with the reverse-proxy-confs removal of the same - which explains it all.

Know what to look for next time around now, thanks for the help!

nemchik:
Ref: https://github.com/linuxserver/reverse-proxy-confs/pull/201 https://github.com/linuxserver/reverse-proxy-confs/pull/200 https://github.com/linuxserver/reverse-proxy-confs/pull/199

nemchik:
For future users:

I've edited the subject, and in pinned this issue so it will be visible even while closed, and I'm locking discussion after posting this.

Proxy configs with web sockets are now handled differently. Users need to manually update their proxy.conf and nginx.conf. See the SWAG readme for more info. https://github.com/linuxserver/docker-swag#updating-configs

We do not need PRs to add back the web socket lines that were removed.

Thanks.

PR #210, opened 2020-09-27
Collabora: Set extra headers for websocket URIs

ryester27:
Was not able to get Collabora working with Nextcloud without these edits


Roxedus:
Please note the new way we handle websockets, as per #198

ryester27:
Sorry. Just noticed you already had a PR open with the changes

PR #220, opened 2020-10-07
Update code-server.subdomain.conf.sample

ImjuzCY:
code-server subdomain showing blank page without these two lines


Roxedus:
Please update your proxy and nginx configs as per #198

PR #231, opened 2020-10-22
add proxmox subdomain conf

ksurl:
single node or load balancer for cluster fqdn

### DIFF
--- a/proxmox.subdomain.conf.sample
+++ b/proxmox.subdomain.conf.sample
@@ -1,0 +1,41 @@
+# make sure that your dns has a cname set for proxmox
+
+server {
+    listen 443 ssl;
+    listen [::]:443 ssl;
+
+    server_name proxmox.*;
+
+    include /config/nginx/ssl.conf;
+
+    client_max_body_size 0;
+
+    # enable for ldap auth, fill in ldap details in ldap.conf
+    #include /config/nginx/ldap.conf;
+
+    # enable for Authelia
+    #include /config/nginx/authelia-server.conf;
+
+    location / {
+        # enable the next two lines for http auth
+        #auth_basic "Restricted";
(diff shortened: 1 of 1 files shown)


nemchik:
Your pr prompted us to realize we need a note in the readme. That's being done here https://github.com/linuxserver/reverse-proxy-confs/pull/234

We don't want to supply premade samples that publicly expose server management UIs.

Thanks for the submission, sorry to have to shoot it down.

ksurl:
thanks, that's fair. for the record I am not exposing it to the internet. I am running swag on a >1024 port for internal access only and have a separate container for public services

ksurl:
what if we specify allow/deny directives in the config to only allow local network access like so:

```
# change to your local subnet of your host as needed and uncomment the line below
#allow 192.168.1.0/24; 
deny all;
```

set to deny all as a failsafe default config.

I was running 2 containers at first, one for internal and one for external access (only one is portforwarded to 443 and the internal only one was bound to a different port), but came across this feature of nginx and it works fine for me.

nemchik:
I agree that the solution you proposed to limit access is doable, but I still don't feel that it is something we would provide samples for. All of the apps you listed as examples to the contrary are things you can run inside of a docker container, and as such can be run with limited access or impact to someone's internal infrastructure (and of course can also be run with significant access and impact as well). This is not an ironclad argument for those samples being provided and for server management UIs to be excluded from consideration. A stronger argument is that we don't prevent you from running custom configs. A sample being absent does not restrict you from adding a config to meet your personal needs. As such, anyone who wants to proxy their server management UIs can do so, and we even provide a proxy sample template which can be used to create a working proxy for those scenarios.

ksurl:
alright thanks.

PR #236, opened 2020-10-26
fix config

wjhrdy:
This config was incomplete. 
The api path is needed for the server to be accessible.
Otherwise you run into problems like this.
https://old.reddit.com/r/homeassistant/comments/6rgf04/frontend_unable_to_connect/

### DIFF
--- a/homeassistant.subdomain.conf.sample
+++ b/homeassistant.subdomain.conf.sample
@@ -39,6 +39,7 @@ server {
     
     location /api {
         include /config/nginx/proxy.conf;
+        resolver 127.0.0.11 valid=30s;
         set $upstream_app homeassistant;
         set $upstream_port 8123;
         set $upstream_proto http;
--- a/homeassistant.subdomain.conf.sample
+++ b/homeassistant.subdomain.conf.sample
@@ -38,9 +38,7 @@ server {
     }
     
     location /api {
-        proxy_http_version 1.1;
-        proxy_set_header Upgrade $http_upgrade;
-        proxy_set_header Connection "upgrade";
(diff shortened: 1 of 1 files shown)


aptalca:
Hmm, sad that the linked report is from 3 years ago, yet no one notified us on any of our channels. Unfortunately we don't prowl reddit for bugs. 

Thanks so much for the PR

homeassistant.subdomain.conf.sample
```diff
@@ -36,4 +36,14 @@ server {
         proxy_pass $upstream_proto://$upstream_app:$upstream_port;
 
     }
+    
+    location /api {
+        proxy_http_version 1.1;
```
aptalca, line 41: Can you remove these 3 lines as they are already handled in the proxy.conf?
https://github.com/linuxserver/docker-swag/blob/master/root/defaults/proxy.conf
Thanks

aptalca: And include proxy.conf instead like above?

homeassistant.subdomain.conf.sample
```diff
@@ -36,4 +36,12 @@ server {
         proxy_pass $upstream_proto://$upstream_app:$upstream_port;
 
     }
+    
+    location /api {
+        include /config/nginx/proxy.conf;
```
nemchik, line 41:
```suggestion
        include /config/nginx/proxy.conf;
        resolver 127.0.0.11 valid=30s;
```

nemchik (changes requested):

nemchik (approved):

PR #238, opened 2020-10-28
update synclounge confs for v3+

aptalca:


drizuid (approved):
I didn't test as I dont use it, but the config seems logical

PR #243, opened 2020-11-29
Add Nextcloud HSTS header

bradleyburgess:
Added the recommended HSTS header from Nextcloud admin panel:

image

### DIFF
--- a/nextcloud.subdomain.conf.sample
+++ b/nextcloud.subdomain.conf.sample
@@ -31,5 +31,6 @@ server {
         proxy_pass $upstream_proto://$upstream_app:$upstream_port;
 
         proxy_max_temp_file_size 2048m;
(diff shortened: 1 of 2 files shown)


aptalca:
Thanks, but it's provided as an option in both proxy.conf in swag, and the default site conf in nextcloud. Both are off by default because hsts is a sledgehammer

bradleyburgess:
@aptalca Thanks! I did check `proxy.conf` but didn't find it in there. I dug around and it's actually in `ssl.conf`. Perhaps a comment in the nextcloud proxy conf pointing to that option?

aptalca:
It's in the nextcloud default site conf with a huge disclaimer: https://github.com/linuxserver/docker-nextcloud/blob/master/root/defaults/default#L17-L26

bradleyburgess:
Gotcha. Thanks.

PR #248, opened 2020-12-09
add version date to proxy confs

aptalca:


nemchik (changes requested):
This is awesome so far.

Let's create a new workflow file in https://github.com/linuxserver/reverse-proxy-confs/tree/master/.github/workflows that will loop all the files and verify that the first line contains what we need in the correct format.

nemchik (approved):

PR #266, opened 2021-01-19
Update znc.subdomain.conf.sample

3n8:
I've successfully port forwarded the IRC port for znc but this won't work without the addition of the stream section in nginx.conf
To be very clear this addition is not for http but to proxy the irc port through swag.
Nginx will accept port 6502 on any subdomain and pass it through to znc, My reasoning for doing it is to avoid configuring ssl certificates in znc.
https://wiki.znc.in/Reverse_Proxy
https://nginx.org/en/docs/stream/ngx_stream_core_module.html

For more info see: https://github.com/linuxserver/docker-swag/pull/58

### DIFF
--- a/znc.subdomain.conf.sample
+++ b/znc.subdomain.conf.sample
@@ -1,6 +1,8 @@
 ## Version 2020/12/09
 # make sure that your dns has a cname set for znc
-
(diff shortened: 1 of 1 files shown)


3n8:
i realised i made a mistake, will get back to it later.

PR #267, opened 2021-01-19
Znc stream

3n8:
Added a few lines of comments to go with: https://github.com/linuxserver/docker-swag/pull/59

the same lines could be added to the znc subfolder .conf but the irc clients i've tried won't work if you specify the subfolder in the client but as the stream is higher in the hierarchy it is not relevant.


nemchik (commented):
Comments made on the linked PR here https://github.com/linuxserver/docker-swag/pull/59#issuecomment-776734684

PR #269, opened 2021-01-29
Handle Nginx http auth + grafana auth separately

draco2003:
Went down a rabbit hole because I had my "admin" type subdirectory protected by http auth, but the individual apps have their own user auth. The passed authorization header was causing problems with Grafana.

### DIFF
--- a/grafana.subdomain.conf.sample
+++ b/grafana.subdomain.conf.sample
@@ -1,4 +1,4 @@
-## Version 2020/12/09
+## Version 2021/01/29
 # make sure that your dns has a cname set for grafana and that your grafana container is not using a base url
 
 server {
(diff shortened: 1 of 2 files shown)


Roxedus (changes requested):
Please update the date at the top

Roxedus:
Would be awesome if you did the same to the subdomain

draco2003:
This passes the build regex:
## Version 2021/01/28
this doesn't
## Version 2021/01/29

Should have posted this yesterday or Monday it looks like ;)

aptalca:
lol, I'll blame @nemchik as I have no energy to try and dissect that massive regex 😅

nemchik:
The other pr wasn't meant to close this, just fix the gha date issue.

nemchik:
@draco2003 please merge from master to get the updated gha and your checks should pass.

Roxedus (approved):

PR #270, opened 2021-01-29
Update pixelfed.subdomain.conf.sample

criadoperez:
pixelfed realeased today the latest version 0.10.10 with default docker and docker-compose configuration.
Using their default setup the container ame is "pixelfed_app_1" not just "pixelfed".
The subdomain pixelfed can stay the same.


Roxedus:
This is easily fixed in the compose-file itself, we will not adapt do this namingscheme, even though it stems from their example.

PR #280, opened 2021-02-02
Create gotify.subfolder.conf.sample

STaRDoGG:
Subfolder RP settings for Gotify

### DIFF
--- a/gotify.subfolder.conf.sample
+++ b/gotify.subfolder.conf.sample
@@ -23,7 +23,4 @@ location /gotify/ {
     set $upstream_port 80;
     set $upstream_proto http;
     proxy_pass $upstream_proto://$upstream_app:$upstream_port;
-
-    proxy_set_header Upgrade $http_upgrade;
-    proxy_set_header Connection "upgrade";
 }
--- a/gotify.subfolder.conf.sample
+++ b/gotify.subfolder.conf.sample
@@ -1,0 +1,29 @@
+## Version 2021/02/02
+# gotify doesn't require a base url
+
+location /gotify {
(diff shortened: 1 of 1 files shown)


gotify.subfolder.conf.sample
```diff
@@ -0,0 +1,29 @@
...
+
+    include /config/nginx/proxy.conf;
+    resolver 127.0.0.11 valid=30s;
+    set $upstream_app gotify;
+    set $upstream_port 80;
+    set $upstream_proto http;
+    proxy_pass $upstream_proto://$upstream_app:$upstream_port;
+
+    proxy_set_header Upgrade $http_upgrade;
+    proxy_set_header Connection "upgrade";
```
Roxedus, line 28: These are redundant.

nemchik: @STaRDoGG This still needs to be addressed (removed). These lines are covered by the proxy.conf https://github.com/linuxserver/docker-swag/blob/d7ce59b839991445e9894de7b53e1ff27134bc42/root/defaults/proxy.conf#L22-L25

STaRDoGG: Again, sorry  about the git newbness, (I only rarely use Git, and only for personal purposes, so I have next to no experience with it); not sure if I did it right. I clicked "Edit", removed lines 27 and 28, committed directly to the patch instead of a new fork and committed changes, but don't see the change reflected here. Did I do it right?

nemchik (approved):

gotify.subfolder.conf.sample
nemchik: Looks good now. Thanks!

PR #312, opened 2021-04-09
Add additional headers for quassel-web

GabrielNagy:
These are taken from the quassel-webserver readme[1]. Without them, the client would fail to connect to the core with messages like:

    Unchecked runtime.lastError: The message port closed before a response was received.

n.b. I was only able to test this with the subfolder configuration, but I added the headers to both configurations for consistency.

[1] https://github.com/magne4000/quassel-webserver/blob/8e3201fa1f6733ae77a351b3694e33b0906a2728/README.md#nginx

### DIFF
--- a/quassel-web.subdomain.conf.sample
+++ b/quassel-web.subdomain.conf.sample
@@ -37,5 +37,11 @@ server {
         set $upstream_proto http;
         proxy_pass $upstream_proto://$upstream_app:$upstream_port;
 
+        proxy_set_header Upgrade $http_upgrade;
+        proxy_set_header Connection "upgrade";
+        proxy_set_header X-Real-IP $remote_addr;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_set_header Host $host;
+        proxy_redirect off;
     }
 }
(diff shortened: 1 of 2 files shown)


quassel-web.subdomain.conf.sample
```diff
@@ -37,5 +37,11 @@ server {
         set $upstream_proto http;
         proxy_pass $upstream_proto://$upstream_app:$upstream_port;
 
+        proxy_set_header Upgrade $http_upgrade;
+        proxy_set_header Connection "upgrade";
+        proxy_set_header X-Real-IP $remote_addr;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_set_header Host $host;
```
Roxedus, line 44: All these lines are redundant, most of them are handled in proxy.conf. The one on the top is handled in nginx.conf, are your files up to date? the swag logs should tell you.

Roxedus (changes requested):

quassel-web.subdomain.conf.sample
GabrielNagy: Hmm, yeah, sorry about this. Looks like I was about 2 years behind on the configs 😅. Something in the updated config files seems to have kicked quassel off, as now it works without any changes. Thanks for the help and sorry again for the noise.

PR #313, opened 2021-04-13
Update nextcloud.subdomain.conf.sample

STaRDoGG:
Added fixes for several common errors seen in the Admin Overview area within Nextcloud, plus the ability to allow loading NC within Organizr, if desired.

### DIFF
--- a/nextcloud.subdomain.conf.sample
+++ b/nextcloud.subdomain.conf.sample
@@ -12,6 +12,13 @@
 #    0 => '192.168.0.1:444', # This line may look different on your setup, don't modify it.
 #    1 => 'nextcloud.your-domain.com',
 #  ),
+#
+# To finish resolving the webfinger/nodeinfo warnings in the overview, edit /config/www/nextcloud/.htaccess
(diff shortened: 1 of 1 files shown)


aptalca:
These are unnecessary as they are all in the Nextcloud nginx confs

STaRDoGG:
> These are unnecessary as they are all in the Nextcloud nginx confs

@aptalca hmmm, I see it now. I wasn't aware of the settings being there. For some reason I still had to add them to my rp conf as I was still getting the warnings in the overview.

The one currently in the nc conf:
`$scheme://$host:$server_port/remote.php/dav;`

is a bit diff from the one I have to use, and I think it's because I have to always append a custom port to my domain. (We've talked about it before, but you might not remember).

Mine has to look like:
`$scheme://$http_host/remote.php/dav;`

but whenever I submit any proxy confs I just remove the `http_` to comply with the requirement.

Still though, I don't see anything in the nc conf for webfinger and nodeinfo, which also need the rewrites added to nc's `.htaccess`.

PR #334, opened 2021-05-30
working homeassistant.subdomain.conf config file

foormea:
1. Some headers seem to always be necessary.
2. I also added a (commented out) way to configure this module, as on my side, I don't want my HA instance to be reachable solely over https -- I also want to be able to reach the instance locally even when the outside internet connection is down.

I'm not aware of any bug open for these, however I've seen a number of people in forums struggling with this.

### DIFF
--- a/homeassistant.subdomain.conf.sample
+++ b/homeassistant.subdomain.conf.sample
@@ -1,4 +1,4 @@
-## Version 2021/05/18
+## Version 2021-05-30
 # make sure that your dns has a cname set for homeassistant and that your homeassistant container is not using a base url
 
 server {
@@ -16,6 +16,9 @@ server {
 
     # enable for Authelia
     #include /config/nginx/authelia-server.conf;
+    
+    # in case the service is running on `host` networking
+    # set $upstream_IP <your.HA.instance.IP>
 
     location / {
         # enable the next two lines for http auth
@@ -35,7 +38,14 @@ server {
         set $upstream_port 8123;
         set $upstream_proto http;
         proxy_pass $upstream_proto://$upstream_app:$upstream_port;
+        
+        # in case the service is running on `host` networking, use this instead:
+        # proxy_pass $upstream_proto://$upstream_IP:$upstream_port;
 
+        # these seem to be needed:
+        proxy_set_header Host $host;
+        proxy_set_header Upgrade $http_upgrade;
+        proxy_set_header Connection "upgrade";
     }
     
     location /api {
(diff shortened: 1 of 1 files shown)


kerta1n:
if you use containerized, home assistant is available from internal access, you just visit the local ip with :8123

foormea:
@kerta1n sure, but the idea is to have Home Assistant reachable from the outside via https, AND internally via IP:8123 in case there's no more working internet connection.

tobbenb:
Maybe I'm missing something, but using IP:port doesn't have anything to do with the reverse proxy. You are then accessing the home assistant container directly. 
If your router supports hairpinning you should be able to access it using the domain also even though your internet is down.

homeassistant.subdomain.conf.sample
```diff
@@ -35,7 +38,14 @@ server {
...
         set $upstream_proto http;
         proxy_pass $upstream_proto://$upstream_app:$upstream_port;
+        
+        # in case the service is running on `host` networking, use this instead:
+        # proxy_pass $upstream_proto://$upstream_IP:$upstream_port;
 
+        # these seem to be needed:
+        proxy_set_header Host $host;
+        proxy_set_header Upgrade $http_upgrade;
+        proxy_set_header Connection "upgrade";
```
Roxedus, line 48: These are not needed. Websockets are handled properly, with proxy.conf and nginx.conf

Roxedus (changes requested):

homeassistant.subdomain.conf.sample
foormea: You're right, `proxy_set_header Host $host;` is already handled, but HA will fail on my side if I don't have the other 2 lines.

Roxedus: > but HA will fail on my side if I don't have the other 2 lines.

Sounds like the files I mentioned are out of date in your instance.

foormea:
@tobbenb The idea behind the commented lines is this: use `network_mode: host` for the HA container, and use swag to connect to it via `<local IP>:8123`. This way HA is reachable on the LAN even when there is no internet, and it's also reachable from the outside on https. The HA container won't be reachable on the LAN if the container is in its own bridged network.

homeassistant.subdomain.conf.sample
foormea: Woops, let me look at that... Sorry for the noise if that's the case!

foormea: Yep, my config files were just not up-to-date... sorry for the noise, will close this. The other aspect of the PR (with the commented lines) doesn't matter that much I guess. Thanks!

PR #341, opened 2021-06-09
Update the Adguard example to encompass first run

fl0r1s:
On first run, the forward should go to port 3000 instead of port 80. After installing, port 80 is the correct one.


aptalca:
This should be left as is. Exposing initial wizard on port 3000 publicly via reverse proxy is not a good idea. One should map port 3000 on first start of adguard and configure it locally, and then the proxy conf will work

Issue #239, opened 2020-11-07
nextcloud not working properly


mzramna:
[linuxserverurl]: https://linuxserver.io
[![linuxserver.io](https://raw.githubusercontent.com/linuxserver/docker-templates/master/linuxserver.io/img/linuxserver_medium.png)][linuxserverurl]
	

raspian

started the letsencrypt with working options and heimdall,(with everyting except next cloud works fine so, i supose it is properly setup) and add nextcloud subfolder,ps: same happends into letsencrypt,nginx and swag,so the prolem should be here

`nginx: [emerg] "proxy_max_temp_file_size" directive invalid value in /config/nginx/proxy-confs/nextcloud.subfolder.conf:35`

using volume mount

just by commenting out the line 35 evryting works just fine,but seems that should be used,accord to nextcloud forums and other referencies,so could be missing any app or anything inside container

cravas:
It's strange that there's no reply about this:
Just change the default value from 2048m to 1024m

Same goes for "nextcloud.subdomain.conf"
armhf doesn't like that high value, ;)

ProgramComputer:
Opened pull request #343 to change to default.

PR #343, opened 2021-06-15
Change to default value for proxy_max_temp_file_size

ProgramComputer:
Received proxy_max_temp_file_size invalid. Checked http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_max_temp_file_size and 1024 is the default.

Fixes issue #239 by changing proxy_max_temp_file_size to default value.


aptalca:
Thanks for the PR, but the default value has nothing to do with this config and it works just fine on all of our test systems. In any case, it is completely user configurable.

PR #352, opened 2021-06-30
Added fritzbox.subdomain.conf

AndreaPro:
Created a proxy-conf to access the Web Interface of Fritzbox routers

### DIFF
--- a/fritzbox.subdomain.conf.sample
+++ b/fritzbox.subdomain.conf.sample
@@ -1,0 +1,47 @@
+## Version 2021/06/30
+# In your router interface under "Home Network > Network > Network Settings > Additional Settings" in the DNS rebind protection  
+# on two separate lines add your domain name and your subdomain + domain name
+#
+# Example:
+# mydomain.com
+# subdomain.mydomain.com
+# 
+# If you have set a LocalDNS like Pi-Hole replace "fritz.box" for $upstream_app with the IP address of your router
+
+server {
+    listen 443 ssl;
+    listen [::]:443 ssl;
+
+    server_name fritzbox.*;
+
+    include /config/nginx/ssl.conf;
+
+    client_max_body_size 0;
+
+    # enable for ldap auth, fill in ldap details in ldap.conf
+    #include /config/nginx/ldap.conf;
+
+    # enable for Authelia
+    #include /config/nginx/authelia-server.conf;
+
+    location / {
+        # enable the next two lines for http auth
+        #auth_basic "Restricted";
+        #auth_basic_user_file /config/nginx/.htpasswd;
+
+        # enable the next two lines for ldap auth
+        #auth_request /auth;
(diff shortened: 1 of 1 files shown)


Roxedus:
> NOTE: We avoid providing samples that publicly expose server management software (ex: syno, qnap, unraid, proxmox, esxi, etc). Pull requests to add samples for this category of applications will not be accepted.

Putting a router on the internet is not a good idea, as it can be dangerous if done wrong.

Closing as this provides great threat if compromised. 

(I know we have exeptions to this rule in the repo already, but ie. unifi, but they have a whole system for authentication, other than just basic_auth. I will reconsider if fritzbox handles authentication better than just with basic_auth)

AndreaPro:
> > NOTE: We avoid providing samples that publicly expose server management software (ex: syno, qnap, unraid, proxmox, esxi, etc). Pull requests to add samples for this category of applications will not be accepted.
> 
> Putting a router on the internet is not a good idea, as it can be dangerous if done wrong.
> 
> Closing as this provides great threat if compromised.
> 
> (I know we have exeptions to this rule in the repo already, but ie. unifi, but they have a whole system for authentication, other than just basic_auth. I will reconsider if fritzbox handles authentication better than just with basic_auth)

Inside the router is possible to enable access from the internet, by reverse proxying it there is the advantage of having a valid certificate for https and on my local configuration I've restricted access to the web interface only if it comes from the local network so I can use it behind a VPN. I have two other files one for OpenMediaVault and one for Cockpit, I can submit them or even these are too dangerous?

TheSpad:
> 
> 
> > > NOTE: We avoid providing samples that publicly expose server management software (ex: syno, qnap, unraid, proxmox, esxi, etc). Pull requests to add samples for this category of applications will not be accepted.
> > 
> > 
> > Putting a router on the internet is not a good idea, as it can be dangerous if done wrong.
> > Closing as this provides great threat if compromised.
> > (I know we have exeptions to this rule in the repo already, but ie. unifi, but they have a whole system for authentication, other than just basic_auth. I will reconsider if fritzbox handles authentication better than just with basic_auth)
> 
> Inside the router is possible to enable access from the internet, by reverse proxying it there is the advantage of having a valid certificate for https and on my local configuration I've restricted access to the web interface only if it comes from the local network so I can use it behind a VPN. I have two other files one for OpenMediaVault and one for Cockpit, I can submit them or even these are too dangerous?

They would also be rejected.

If you know what you're doing you can obviously setup whatever confs you want to, but we don't want to publish any sample confs that could lead to people accidentally exposing their infrastructure management interfaces to the internet.

Roxedus:
Those falls under the same category. 

> __*I've*__ restricted access to the web interface only if it comes from the local network

Most people will just enable it, and be unprotected. Our stance here is; if you know how to make it safe, you know how to create the proxy

PR #363, opened 2021-07-10
Allow Home Assistant local files to bypass auth

golles:
I've enabled Authalia myself and noticed that push notifications containing images from the `/local` (www) directory, wouldn't load (which makes sense as there was no auth).

Let me know if this is acceptable or if you would rather merge`api` and `local` rule with a regex, my personal preference is this exact match.

### DIFF
--- a/homeassistant.subdomain.conf.sample
+++ b/homeassistant.subdomain.conf.sample
@@ -46,4 +46,13 @@ server {
         set $upstream_proto http;
         proxy_pass $upstream_proto://$upstream_app:$upstream_port;
     }
+    
+    location /local {
+        include /config/nginx/proxy.conf;
+        include /config/nginx/resolver.conf;
+        set $upstream_app homeassistant;
+        set $upstream_port 8123;
+        set $upstream_proto http;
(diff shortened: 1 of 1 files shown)


aptalca:
Separate should be fine.

Is there anything sensitive served from the local folder?

I do use Authelia with ha but don't use push notifications. How are the notifications set up?

golles:
> Separate should be fine.
> 
> Is there anything sensitive served from the local folder?
> 
> I do use Authelia with ha but don't use push notifications. How are the notifications set up?

There shouldn't be any sensitive data in that dir, this is meant to put images and javascript files to be used in your frontend and push notifications (https://companion.home-assistant.io/docs/notifications/notification-attachments/#www-folder).
This folder is also not protected by the standard Home Assistant auth.

Having said all of this, it's the user's own responsibility to not expose sensitive items in the www folder, but that counts for any system exposed to the internet.

For more information about push notifications is best to check the documentation on the earlier mentioned https://companion.home-assistant.io website, they put a good effort into making it easy to understand.

aptalca:
Cool, if it's not protected by ha's built in auth, that's good enough for me

aptalca (approved):

PR #364, opened 2021-07-11
Bypass Authelia auth for Calibre-Web OPDS feed

Daniel15:
eBook apps do not support fancy form-based authentication nor two-factor auth, like what's provided by Authelia. This PR updates the config to bypass Authelia for the `/opds/...` endpoints. I verified that this allows me to use Calibre-Web's OPDS feed with Moon+ Reader on Android.

I also added a comment about how to enable Authelia single sign-on support in Calibre-Web. Let me know if you'd prefer this to be mentioned in the docs instead.

### DIFF
--- a/calibre-web.subdomain.conf.sample
+++ b/calibre-web.subdomain.conf.sample
@@ -27,6 +27,9 @@ server {
         #error_page 401 =200 /ldaplogin;
 
         # enable for Authelia
+        # To use Authelia to log in to Calibre-Web, make sure "Reverse Proxy Login" is 
+        # enabled, "Reverse Proxy Header Name" is set to Remote-User, and each Authelia
+        # user also has a corresponding user manually created in Calibre-Web.
         #include /config/nginx/authelia-location.conf;
 
         include /config/nginx/proxy.conf;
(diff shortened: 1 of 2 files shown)


Roxedus:
A few questions:
The endpoint only exposes books?
Does this endpoint have a way to list files? (some people may not want to braodcast their library to everyone)

Daniel15:
@Roxedus 

> The endpoint only exposes books?

It exposes any content loaded into Calibre-web (books, magazines, and audiobooks)

> Does this endpoint have a way to list files? (some people may not want to braodcast their library to everyone)

Anonymous (unauthenticated) access is disabled by default, so you need to authenticate with a username and password configured in calibre-web itself. To allow anyone to access it, you need to explicitly enable anonymous access in the calibre-web settings (which could be useful if the library consists entirely of public domain books, or it's a local installation that's only accessible on the local network, or via a VPN).

So, when you have Authelia enabled:
- When you hit Calibre-web in a browser, it relies entirely on Authelia. It'll redirect to Authelia if you're not logged in yet, and if you are logged in, it'll automatically log in to Calibre-web with the same username, assuming the "Reverse Proxy Login feature" is enabled
- When you hit Calibre-web via an API client that does not support Authelia (every OPDS client), it'll rely on HTTP Basic authentication, using the usernames and passwords configured directly in Calibre-web

When you do not have Authelia enabled:
- All auth (both through Calibre-web Web UI and via the API) is handled via Calibre

I'd like if Authelia supported HTTP Basic auth in addition to form-based auth for API use cases, but it's tricky since there's no way to do two-factor auth using Basic auth, so I can't think of a useful way of doing it. So for now, Authelia is only useful for web UIs, not for APIs.

Roxedus (approved):

Roxedus:
> Anonymous (unauthenticated) access is disabled by default

That's enough for me :)

PR #367, opened 2021-07-14
Fixes for NC21 - changes /.well-known block

Wdavery:
Before this change the config results in a warning `Your web server is not properly set up to resolve "/.well-known/webfinger".`

Source of proposed fix: https://github.com/linuxserver/docker-nextcloud/issues/189#issuecomment-846376583


tobbenb (approved):

tobbenb:
Thanks for this! I kept forgetting it.

PR #369, opened 2021-07-25
Add additional location for /confirmdeletion

mrskizzex:
Adding location for /confirmdeletion for better compatibility with Authelia.

### DIFF
--- a/qbittorrent.subdomain.conf.sample
+++ b/qbittorrent.subdomain.conf.sample
@@ -1,4 +1,4 @@
-## Version 2021/05/18
+## Version 2021/07/25
 # make sure that your dns has a cname set for qbittorrent and that your qbittorrent container is not using a base url
 
 server {
@@ -97,6 +97,20 @@ server {
     }
 
     location ~ (/qbittorrent)?/sync {
+        include /config/nginx/proxy.conf;
+        include /config/nginx/resolver.conf;
+        set $upstream_app qbittorrent;
+        set $upstream_port 8080;
+        set $upstream_proto http;
+        proxy_pass $upstream_proto://$upstream_app:$upstream_port;
+
+        rewrite /qbittorrent(.*) $1 break;
+
+        proxy_set_header Referer '';
+        proxy_set_header Host $upstream_app:$upstream_port;
+    }
+    
+    location ~ (/qbittorrent)?/confirmdeletion {
         include /config/nginx/proxy.conf;
         include /config/nginx/resolver.conf;
         set $upstream_app qbittorrent;


aptalca:
We don't want to bypass auth for deletion

mrskizzex:
I made this suggestion after encountering errors with default setup.
Deleting one torrent in qBittorent works but two or more redirects to Authelia.

```
XX.XX.XX.XX - - [25/Jul/2021:15:16:42 +0200] "GET /confirmdeletion.html?hashes=2383b50f94801bf93d700228f0e13c13884853a1&deleteFiles=false HTTP/2.0" 200 791 "https://qbtvpn.example.com/"
```
Returns 200 and shows deletion confirmation

```
XX.XX.XX.XX - - [25/Jul/2021:15:16:50 +0200] "GET /confirmdeletion.html?hashes=8042fc9c4910278aa330215aeede403e9c84a42f|bb1f3bc4cd20e514b9867317e463ddb746e57ca5&deleteFiles=false HTTP/2.0" 302 138 "https://qbtvpn.example.com/" 
XX.XX.XX.XX - - [25/Jul/2021:15:16:50 +0200] "GET /authelia/?rd=https://qbtvpn.example.com/confirmdeletion.html?hashes=8042fc9c4910278aa330215aeede403e9c84a42f|bb1f3bc4cd20e514b9867317e463ddb746e57ca5&deleteFiles=false HTTP/2.0" 200 506 "https://qbtvpn.example.com/"
```
Returns 302 and shows Authelia

There always is qBittorrent internal auth so I'm unsure if this has big impact on anything.

aptalca:
Try commenting out these three lines:
https://github.com/linuxserver/docker-swag/blob/master/root/defaults/authelia-server.conf#L13-L15

mrskizzex:
That works but aren't they supposed to be there?

> We don't want to bypass auth for deletion
I was investigated this a bit and you are already doing it, in fact the default file is already exposing underlying api completely and only qbittorent auth is stopping it.  The only thing my addition is doing is returing this box to the user if someone entered 
https://qbtvpn.example.com/confirmdeletion.html?hashes=2383b50f94801bf93d700228f0e13c13884853a1&deleteFiles=false
into their browser. Buttons don't work at all because they require whole page to be loaded properly to work.
image

mrskizzex:
Well my fix doesn't work as well as I thought, this pr can be scrapped. Your fix fixes the root of the issue.
@TheSpad said on discord that those 3 lines were added to mitigate CVE in older versions of authelia, maybe it's worth mentioning in that file? So users can uncomment those out if they have older versions.

PR #375, opened 2021-09-05
Add proxy_set_header X-Scheme https;

Starker3:
The X-scheme header seems to be required for Octoprint to correctly use Websockets.

### DIFF
--- a/octoprint.subdomain.conf.sample
+++ b/octoprint.subdomain.conf.sample
@@ -1,4 +1,4 @@
-## Version 2021/06/16
+## Version 2021/09/05
 # make sure that your dns has a cname set for octoprint and that your octoprint container is not using a base url
 
 server {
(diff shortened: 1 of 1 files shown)


nemchik:
I've looked a few times since seeing this and I can't find consistent information supporting the need/use for this header.

Does octoprint have documentation about this? The headers we have in swag proxy.conf should universally allow websockets.

Starker3:
@nemchik This was documented by the Project Lead for Octoprint in the forum here: https://community.octoprint.org/t/octoprint-doesnt-work-when-connecting-via-https/225

The default proxy.conf doesn't seem to work for Octoprint, it possible to fix this in Octoprint's configuration, but its not something that is enabled by default.

nemchik (approved):

PR #378, opened 2021-09-12
Auto redirect for non-standard port

wswv:
Purpose to add function to auto redirect to non-standard port when server works on non-standard port condition.

### DIFF
--- a/calibre-web.subdomain.conf.sample
+++ b/calibre-web.subdomain.conf.sample
@@ -54,5 +54,6 @@ server {
         set $upstream_proto http;
         proxy_pass $upstream_proto://$upstream_app:$upstream_port;
         proxy_set_header X-Scheme $scheme;
+	proxy_redirect https://$host/ https://$host:selfdefined port number/;#here use your self-defined port number, so it could redirect to other link then still keep self defined port number.
     }
 }


wswv (commented):
test passed in my server

calibre-web.subdomain.conf.sample
```diff
@@ -54,5 +54,6 @@ server {
         set $upstream_proto http;
         proxy_pass $upstream_proto://$upstream_app:$upstream_port;
         proxy_set_header X-Scheme $scheme;
+	proxy_redirect https://$host/ https://$host:selfdefined port number/;#here use your self-defined port number, so it could redirect to other link then still keep self defined port number.
```
nemchik, line 57:
```suggestion
        proxy_redirect https://$host/ https://$host:selfdefined port number/;#here use your self-defined port number, so it could redirect to other link then still keep self defined port number.
```
Matching indent

nemchik (changes requested):

nemchik:
Also, can you set the upstream port variable and then use that variable in your line?

Ex:
```nginx
        set $upstream_port 8444;
...
	proxy_redirect https://$host/ https://$host:$upstream_port/;
```

nemchik:
Lastly, I think we might want this line commented by default. I'm thinking not everyone will be wanting do forward to a custom port in this way.

drizuid:
OP didn't make the adjustments requested and swag has been revamped requiring a re-work. closing this.

PR #383, opened 2021-10-02
Added config sample for the Synapse Matrix Server

deutschepost:

Synapse provides a default nginx configuration [here](https://github.com/matrix-org/synapse/blob/master/docs/reverse_proxy.md). I merged it with the template.

##  Thanks, team linuxserver.io


synapse.subdomain.conf.sample
```diff
@@ -0,0 +1,44 @@
+server {
```
nemchik, line 1: Please include a version line with today's date. See https://github.com/linuxserver/reverse-proxy-confs/blob/4ef5a778680461a85050d1bbaa3aca08ac1470ca/_template.subdomain.conf.sample#L1

synapse.subdomain.conf.sample
```diff
@@ -0,0 +1,44 @@
...
+        include /config/nginx/proxy.conf;
+        include /config/nginx/resolver.conf;
+        set $upstream_app synapse;
+        set $upstream_port 8008;
+        set $upstream_proto http;
+        proxy_pass $upstream_proto://$upstream_app:$upstream_port;
+
+        proxy_set_header X-Forwarded-For $remote_addr;
+        proxy_set_header X-Forwarded-Proto $scheme;
+        proxy_set_header Host $host;
```
nemchik, line 42: Does it work without these? They are included in the proxy.conf here https://github.com/linuxserver/docker-swag/blob/ce33eeebe79e3e543a5c3b97ee2a1914d71d825f/root/defaults/proxy.conf#L21-L30 but not all of the values are the same. The ones we have as default in proxy.conf should be pretty universal.

synapse.subdomain.conf.sample
```diff
@@ -0,0 +1,44 @@
...
+
+    client_max_body_size 0;
+
+    # enable for ldap auth, fill in ldap details in ldap.conf
+    #include /config/nginx/ldap.conf;
+
+    # enable for Authelia
+    #include /config/nginx/authelia-server.conf;
+
+    location ~* ^(\/_matrix|\/_synapse\/client) {
```
nemchik, line 21: With this, would you have to use `matrix.domain.com/_matrix` or `matrix.domain.com/_synapse/client` ? would `matrix.domain.com/` not work?

nemchik (approved):

synapse.subdomain.conf.sample
deutschepost: In the [documentation](https://github.com/matrix-org/synapse/blob/master/docs/reverse_proxy.md) I referenced. They explain, that these URLs should be accessible through the Reverse Proxy. I'm not quite shure if it would work without them.

nemchik: Would you give it a try with `location / {`

I think that would work with a subdomain proxy. A subfolder proxy might be different and that location in their docs might make more sense.

If it doesn't work with the basic location I'm suggesting we can keep what you have.

synapse.subdomain.conf.sample
deutschepost: Good Question. The Host one is identical. The X-Forwarded-For in the proxy.conf seems to append $remote_addr. I'm not shure about the Proto. My setup is currently offline to change some configs. I don't know when I would have time to try it out. And also how to check if everything would be working perfectly...

synapse.subdomain.conf.sample
deutschepost: Ok. I'll try it out when my setup is back online.

synapse.subdomain.conf.sample
deutschepost: I removed all of these and it seems to work perfectly

synapse.subdomain.conf.sample
deutschepost: I replaced the location with / and it seems to wirk just fine

deutschepost:
I am actually not sure if the Authorization Stuff really makes sense in this application. The synapse server does not have any web interfaces.

deutschepost:
I dont't understand what the problem is. Could you elaborate, what I have to do, to have this PR merged?

nemchik:
> I dont't understand what the problem is. Could you elaborate, what I have to do, to have this PR merged?

It looks correct to me. I'll have to have a look later today.

CarbonMac:
This should be added to that config for federation via .well-known auth

`  # Edit the following for federation on your matrix server
   # location ^~ /.well-known/matrix/server {
   #     return 200 '{ "m.server": "matrix.subdomain.com443" }';
   # }`

deutschepost:
I think my last commit could have fixed it. Just ran the workflow on my local machine.

nemchik:
I'll add a check to the CI to warn about line endings, and an `.editorconfig` to the repo to hopefully cut down on this in the future.

Great catch, and thanks for finding it!

deutschepost:
Thanks for merging. If I improve something in the future I'll submit another request.

PR #385, opened 2021-10-05
Websockets

symetryczny:
As per https://gotify.net/docs/nginx

### DIFF
--- a/gotify.subdomain.conf.sample
+++ b/gotify.subdomain.conf.sample
@@ -18,6 +18,23 @@ server {
     #include /config/nginx/authelia-server.conf;
 
     location / {
+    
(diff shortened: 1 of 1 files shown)


j0nnymoe:
Have you tested this? Our swag container is setup to automatically detect any websockets needed.

aptalca:
Also many of those are already covered in other nginx confs we provide

symetryczny:
Frankly speaking I did not check if webhooks worked before I amended the reverse proxy config, only after I did the changes. I am closing the request if you guys confirm that they are covered with current config. Will check later and if something doesn't work with the current config I will open another pull request.

PR #399, opened 2021-11-03
modify seafile config

drizuid:
couple issues with this as-is. 
1) SEAFILE_SERVER_LETSENCRYPT=true tells seafile to request a letsencrypt certs which is superfluous if using swag.
2) SEAFILE_SERVER_HOSTNAME= also seems relevant to the cert request process, though it may be needed, so i wont kill it in this PR

it's superfluous to proxy to 443 when 80 is also listening though this has been reported as causing issues previously so sufficient testing is needed.

it was also missing a bunch of our normal shit so this PR bring it in line with the standard.


drizuid:
this will need someone to test thoroughly due to some previously reported issues in NOT proxying to https.

drizuid:
I forgot about this and we revamped swag, closing this.

PR #401, opened 2021-11-09
Update youtube-dl-server.subdomain.conf.sample

onvrb:
fixed to work with the latest version - subfolder is now required


nemchik:
Ref: https://github.com/linuxserver/reverse-proxy-confs/pull/401 https://github.com/linuxserver/reverse-proxy-confs/pull/402

Can you point to the documentation showing where this is required? I didn't see it in the repo.

onvrb:
> Ref: #401 #402
> 
> Can you point to the documentation showing where this is required? I didn't see it in the repo.

I didn't see it on the docs, probably they didn't even exist back then, but that's what it got working. It has been a while ago anyway, so I can't confirm if it still works because I don't use it anymore.

PR #407, opened 2021-12-04
Update nextcloud.subdomain.conf.sample

BrodyStone21:
Upon struggling with getting this figured out for a while, I realized that this is attempting to gain access to Nextcloud through port 443. This is fine with the linuxserver/nextcloud image, but with the official image it won't work as the official image runs on http and port 80.


aptalca:
Thanks for the PR but if there is an image published by us for the app in question, it should be assumed that the proxy conf we provide is for our image.

We don't want to give the impression that we provide support for 3rd party images when we maintain our own.

BrodyStone21:
Completely understand. That was honestly my bad, I forked it and then realized it was linuxserver's repo and not SWAG's...

Appreciate the clarification!

PR #415, opened 2022-01-03
Update the server_name for uptime-kuma subdomain from uptime.* to uptime-kuma.*

catchsudheera:
This is to keep the consistency. I spent some time before realizing that `uptime-kuma `needs a subdomain as only `uptime` this is a bit misleading and effectively prevents someone from using another container in the same network called just `uptime` which nothing to do with `uptime-kuma`. 

Not sure how this would affect the backward compatibility, just thought of updating it in case this was a mistake. plus the original file added very recently making the impact lower.

Thanks

### DIFF
--- a/uptime-kuma.subdomain.conf.sample
+++ b/uptime-kuma.subdomain.conf.sample
@@ -5,7 +5,7 @@ server {
     listen 443 ssl;
     listen [::]:443 ssl;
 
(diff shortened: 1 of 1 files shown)


catchsudheera:
@Starker3 FYI

nemchik (approved):

PR #417, opened 2022-01-14
Fix CORS for web clients

tommytran732:
https://matrix-org.github.io/synapse/latest/setup/installation.html

/.well-known/matrix/client needs Access-Control-Allow-Origin * for web clients to work

### DIFF
--- a/synapse.subdomain.conf.sample
+++ b/synapse.subdomain.conf.sample
@@ -38,5 +38,10 @@ server {
         set $upstream_port 8008;
         set $upstream_proto http;
         proxy_pass $upstream_proto://$upstream_app:$upstream_port;
+        
+    	location /.well-known/matrix/client {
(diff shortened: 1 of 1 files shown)


nemchik:
I notice the page you linked has the following section https://matrix-org.github.io/synapse/latest/setup/installation.html#client-well-known-uri and their example has an additional line that looks like it might not be universally reusable.
```nginx
    return 200 '{"m.homeserver": {"base_url": "https://<matrix.example.com>"}}';
```
Is this line also required? we can consider using an nginx variable in place of `<matrix.example.com>` or leaving the line commented out and instruct users to fill it accordingly?

tommytran732:
I don't think it's needed so long as https://<server_name>/.well-known/matrix/client returns the correct json file.

PR #419, opened 2022-01-18
Added Duplicacy configuration files

morpheus65535:


Roxedus (changes requested):
The 307 and 301 redirects should be consolidated, i would also see if one could do a condition for the returns, like the plex subfolder example

morpheus65535:
> The 307 and 301 redirects should be consolidated, i would also see if one could do a condition for the returns, like the plex subfolder example

I don't see how it can be done since for Plex all the file reside in a `web` subdirectory where duplicacy store it's files at root. Doind the same as plex would mean that every requests for any location would result in a redirect to duplicacy.

Maybe someone else will come with a hint on how to fix this but I don't see any cleaner than what I did.

nemchik (changes requested):

Ref:

duplicacy.subfolder.conf.sample
```diff
@@ -0,0 +1,123 @@
...
+    return 307 $scheme://$host/duplicacy/start_stop_schedule;
+}
+
+location /stop_restore {
+    return 307 $scheme://$host/duplicacy/stop_restore;
+}
+
+location /update_listing_session {
+    return 307 $scheme://$host/duplicacy/update_listing_session;
+}
```
nemchik, line 99: All these locations are not beneath the subfolder (ex: `location /duplicacy/update_listing_session {`)
Does this app actually work with subfolder?

morpheus65535: Yes but it's not perfect. It require to add each and every path in the proxy config and conflict may happens with other software. Not sure if it's robust enough to be included anymore.

nemchik: Let's just keep the subdomain file for now. If you don't mind, delete the subfolder file and i'll review/accept the subdomain conf.

morpheus65535: Done!

nemchik (approved):

PR #430, opened 2022-03-09
Added sample config for linuxserver/grav

ofersadan85:

### DIFF
--- a/grav.subfolder.conf.sample
+++ b/grav.subfolder.conf.sample
@@ -1,0 +1,23 @@
+## Version 2022/03/09
+# In order to use this location block you need to edit the default file one folder up and comment out the / location
+
+location / {
+    # enable the next two lines for http auth
+    #auth_basic "Restricted";
+    #auth_basic_user_file /config/nginx/.htpasswd;
+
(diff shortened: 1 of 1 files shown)


grav.subfolder.conf.sample
```diff
@@ -0,0 +1,23 @@
+## Version 2022/03/09
+# In order to use this location block you need to edit the default file one folder up and comment out the / location
```
Roxedus, line 2: Pretty sure the php location also interferes.

Roxedus (changes requested):
See note

grav.subfolder.conf.sample
ofersadan85: When I tested this (been a while) I remember not touching the php location and it worked fine. Would you be able to test to verify if this is needed @Roxedus ?

ofersadan85:
@Roxedus see my question on your review

Roxedus:
I have no environment handy to test this, i just know it was a issue with the organizr, and possibly the heimdall contaieners, which both are PHP based.

PR #431, opened 2022-03-17
adding headers for guacamole

jonalter:
I setup swag in front of [calibre](https://docs.linuxserver.io/images/docker-calibre) , but it wouldn't show the UI when when I could connect through the reverse proxy. I got it to work after adding these headers which were suggested by the [guacamole docs](https://guacamole.apache.org/doc/0.9.7/gug/proxying-guacamole.html) under the **Proxying Guacamole** section.


aptalca:
These headers are already included in other SWAG confs

jonalter:
So they are. Still not sure why adding them solved my problem. Definitely on my end though. Thanks

PR #442, opened 2022-05-10
Updates for Nextcloud security checks

hollanbm:
Identical changes as #367 but made in the subdomain conf

resolves warnings from Nextcloud security check

### DIFF
--- a/nextcloud.subdomain.conf.sample
+++ b/nextcloud.subdomain.conf.sample
@@ -33,4 +33,21 @@ server {
 
         proxy_max_temp_file_size 2048m;
     }
+    
+    # The rules in this block are an adaptation of the rules
+    # in the Nextcloud `.htaccess` that concern `/.well-known`.
+    location /.well-known/carddav {
+        return 301 $scheme://$host/remote.php/dav;
+    }
+
+    # The rules in this block are an adaptation of the rules
+    # in the Nextcloud `.htaccess` that concern `/.well-known`.
+    location /.well-known/caldav {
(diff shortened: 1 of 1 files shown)


hollanbm:
Should I update the date in the comment at the top, or is that done automatically on build?

Roxedus:
You need to update it

hollanbm:
> You need to update it

done. thanks

aptalca:
This shouldn't be needed as nextcloud's site conf already handles the redirects

hollanbm:
> This shouldn't be needed as nextcloud's site conf already handles the redirects

Am I missing something?

These config options were required for me to pass the security checks in the NextCloud app Admin Overview section

Without
image
  
Without
image

aptalca:
https://github.com/linuxserver/docker-nextcloud/blob/master/root/defaults/default#L88-L101

hollanbm:
> https://github.com/linuxserver/docker-nextcloud/blob/master/root/defaults/default#L88-L101

Thank you. This installation is a few years old. I just keep upgrading in place.I don't see a date at the top of my current `site-conf` in the nextcloud container. But mine is definitely dated.

I'll compare against the other files in the default folder and see what I need to update

aptalca:
If you haven't customized it, just delete it and restart the container. It will generate the latest one in there.

hollanbm:
> If you haven't customized it, just delete it and restart the container. It will generate the latest one in there.

I hadn't, I'll do that. Thanks for the help

PR #449, opened 2022-06-08
update for authentik

bokkoman:
Example for use of Authentik.
Copy paste these to each conf that needs it.


nemchik:
Replaced by https://github.com/linuxserver/reverse-proxy-confs/pull/521

PR #455, opened 2022-06-26
fireflyiii family subdomain sample configs

nomandera:

### DIFF
--- a/fireflyiii-importer.subdomain.conf.sample
+++ b/fireflyiii-importer.subdomain.conf.sample
@@ -1,0 +1,44 @@
+## Version 2022/06/24
+# make sure that your dns has a cname set for fireflyiii-importer and that your fireflyiii-importer container is named fireflyiii-importer
+# fireflyiii-importer will require `TRUSTED_PROXIES=**` variable or similar
+
+server {
+    listen 443 ssl;
+    listen [::]:443 ssl;
+
+    server_name fireflyiii-importer.*;
+
+    include /config/nginx/ssl.conf;
+
+    client_max_body_size 0;
+
+    # enable for ldap auth, fill in ldap details in ldap.conf
+    #include /config/nginx/ldap.conf;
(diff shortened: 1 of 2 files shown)


fireflyiii-importer.subdomain.conf.sample
```diff
@@ -0,0 +1,44 @@
+## Version 2022/06/24
+# make sure that your dns has a cname set for fireflyiii-importer and that your fireflyiii-importer container is named fireflyiii-importer
+# fireflyiii-importer will require `TRUSTED_PROXIES=**` variable or similar
```
Roxedus, line 3: Does it work with setting `TRUSTED_PROXIES` to `swag`. or should we recommend the whole default docker bridge block?

nomandera: According to [documentation](https://docs.firefly-iii.org/firefly-iii/installation/docker/)

`TRUSTED_PROXIES which must be set to either the reverse proxy machine or simply **`

I have made this change and it seems to work with fireflyiii but not fireflyiii-importer.

I will update the PR once I make them both work.

Nice catch.

nomandera: I spoke too soon. Something expired with the `TRUSTED_PROXIES` set to anything other than `**` you can no longer authenticate.

Happy to try anything else we can think of but at this point it looks the only reliable option is the default.

Worth noting the sample .env file makes no reference to any other options

https://github.com/firefly-iii/firefly-iii/blob/main/.env.example#L32-L34

My take on this is that for now at least we just add a line to the PR samples giving context.

nemchik (changes requested):

Ref:

nomandera:
I cannot make this work reliably and am currently porting away from Firefly-iii because of this and other edge case complications.

Closing PR as I will likely never be able to fix it.

PR #464, opened 2022-07-21
Annotate files

Roxedus:
Main feature of this pr is to annotate files which triggers the existing checks.
It also does a bump of the checkout action.
~~linter go brr~~

### DIFF
--- a/.github/workflows/check_samples.yml
+++ b/.github/workflows/check_samples.yml
@@ -18,8 +18,9 @@ jobs:
           NOT_SAMPLES=$(find . -not -path '*/\.*' -type f ! \( -name '*.conf.sample' -o -name 'README.md' -o -name 'LICENSE' \))
           NOT_SAMPLES_COUNT=$(echo "${NOT_SAMPLES}" | wc -w)
           if (( NOT_SAMPLES_COUNT > 0 )); then
-              echo "The following files have extensions that are not allowed:"
-              echo "${NOT_SAMPLES}"
+              for i in ${NOT_SAMPLES}; do
+                echo "::error file=${i},line=1,title=Disallowed filenames::This file extension is not allowed, only .sample is allowed"
+              done
               exit 1
           fi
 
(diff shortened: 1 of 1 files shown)


.github/workflows/check_samples.yml
```diff
@@ -2,54 +2,58 @@ name: Check Samples
...
+      - name: Check Allowed File Names
+        run: |
+          NOT_SAMPLES=$(find . -not -path '*/\.*' -type f ! \( -name '*.conf.sample' -o -name 'README.md' -o -name 'LICENSE' \))
+          NOT_SAMPLES_COUNT=$(echo "${NOT_SAMPLES}" | wc -w)
+          if (( NOT_SAMPLES_COUNT > 0 )); then
+              for i in ${NOT_SAMPLES}; do
+                echo "::error file=${i},line=1,title=Disallowed filenames::This file extension is not allowed, only .sample is allowed"
+              done
+              exit 1
+          fi
```
nemchik, line 25:
```suggestion
          NOT_SAMPLES=$(find . -not -path '*/\.*' -type f ! \( -name '*.conf.sample' -o -name 'README.md' -o -name 'LICENSE' \) -exec grep echo "::error file={},line=1,title=Disallowed filenames::This file extension is not allowed, only .sample is allowed" \;)
          NOT_SAMPLES_COUNT=$(echo "${NOT_SAMPLES}" | wc -w)
          if (( NOT_SAMPLES_COUNT > 0 )); then
            echo "${NOT_SAMPLES}"
            exit 1
          fi
```
This isn't a huge difference. It's slightly cleaner with regards to usage of syntax, but the line length actually makes the complexity feel higher and less maybe readable.

I've only left the comment on this one, but this syntax would apply to all the other changes as well.

Roxedus: This complexity reaches borderline unreadable imo.

nemchik: Sounds good

nemchik (approved):

PR #469, opened 2022-08-11
Change default container name

jwmoss:
The name is changedetection.io in your [docker container on dockerhub](https://hub.docker.com/r/linuxserver/changedetection.io)


drizuid:
it's based on `container_name: changedetection`

PR #470, opened 2022-08-12
add missing semicolon to monica-subdomain example

dhitchcock:
Just added a semicolon that was missing the monica subdomain example.  I copied it and used it and it didn't work, so saving anyone else doing the same!


aptalca (approved):

aptalca:
Thanks for the PR

PR #483, opened 2022-09-08
Adjustments for standard-base

nemchik:
https://github.com/linuxserver/docker-swag/pull/169
https://github.com/linuxserver/docker-baseimage-alpine-nginx/pull/84
https://github.com/linuxserver/docker-mods/pull/502

Merge with swag PR.

Existing user configs should not be affected. Container will show update messages in logs.

### DIFF
--- a/_template.subdomain.conf.sample
+++ b/_template.subdomain.conf.sample
@@ -1,4 +1,4 @@
-## Version 2021/05/18
+## Version 2022/09/08
 # REMOVE THIS LINE BEFORE SUBMITTING: The structure of the file (all of the existing lines) should be kept as close as possible to this template.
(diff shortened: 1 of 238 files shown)


aptalca:
This also requires the auto-proxy mod updated to support both the old and the new proxy confs: https://github.com/linuxserver/docker-mods/blob/swag-auto-proxy/root/app/auto-proxy.sh#L79

nemchik:
> This also requires the auto-proxy mod updated to support both the old and the new proxy confs: https://github.com/linuxserver/docker-mods/blob/swag-auto-proxy/root/app/auto-proxy.sh#L79

Opened https://github.com/linuxserver/docker-mods/pull/502

aptalca (approved):

PR #484, opened 2022-09-14
Create ntfy.subdomain.conf.sample

the-lazy-fox:
Adding ntfy for subdomains


Roxedus (changes requested):
Either explain why this sample does not include auth, or include the default commented basic auth and authelia configs.

Issue #323, opened 2021-04-29
Config request: ProjectSend


ceptonit:
May I request a config for ProjectSend?
https://fleet.linuxserver.io/image?name=linuxserver/projectsend

Best regards.

Taubin:
Does anyone have a working config for this? I've searched high and low and cannot figure out how to proxy this service as a subdomain in swag.

PR #486, opened 2022-09-17
Subfolder usage for projectsend

mcinj:
Addresses #323 

Tested with https://github.com/linuxserver/docker-projectsend


Roxedus (changes requested):
This syntax is invalid

nemchik (changes requested):

Ref:

PR #501, opened 2022-10-28
Fix nextcloud X-Frame-Options warning

nemchik:
In Nextcloud's admin overview.

Before:
image

After:
image

### DIFF
--- a/nextcloud.subdomain.conf.sample
+++ b/nextcloud.subdomain.conf.sample
@@ -1,4 +1,4 @@
-## Version 2021/05/18
+## Version 2022/10/28
 # make sure that your dns has a cname set for nextcloud
 # assuming this container is called "swag", edit your nextcloud container's config
 # located at /config/www/nextcloud/config/config.php and add the following lines before the ");":
@@ -31,6 +31,7 @@ server {
         set $upstream_proto https;
         proxy_pass $upstream_proto://$upstream_app:$upstream_port;
 
+        proxy_hide_header X-Frame-Options;
         proxy_max_temp_file_size 2048m;
     }
 }
--- a/nextcloud.subfolder.conf.sample
+++ b/nextcloud.subfolder.conf.sample
@@ -1,4 +1,4 @@
-## Version 2021/07/14
+## Version 2022/10/28
 # Assuming this container is called "swag", edit your nextcloud container's config
 # located at /config/www/nextcloud/config/config.php and add the following lines before the ");":
 #  'trusted_proxies' => ['swag'],
@@ -32,6 +32,7 @@ location ^~ /nextcloud/ {
     proxy_pass $upstream_proto://$upstream_app:$upstream_port;
 
     rewrite /nextcloud(.*) $1 break;
+    proxy_hide_header X-Frame-Options;
     proxy_max_temp_file_size 2048m;
     proxy_set_header Range $http_range;
     proxy_set_header If-Range $http_if_range;


thespad (approved):

jdancouga:
I still get the same warning as your "Before:" screenshot after this commit update. I end up having to enable "add_header X-Frame-Options "SAMEORIGIN" always" in swag-docker's ssl.conf file to get rid of the warning. Am I doing this right?

nemchik:
> I still get the same warning as your "Before:" screenshot after this commit update. I end up having to enable **add_header X-Frame-Options "SAMEORIGIN" always** in swag-docker's ssl.conf file to get rid of the warning. Am I doing this right?

Yes, you need that as well. It's not universally required in every use case that swag could be used for, but it is ideal to use that setting with nextcloud.

zedocrob:
I've also added `add_header X-Frame-Options "SAMEORIGIN" always` in swag-docker's ssl.conf and disabled
`#proxy_hide_header X-Frame-Options;` in nextcloud.subdomain.conf.
When enabled i get the warning In Nextcloud's admin overview.

cdelledonne:
On my end, "Before" and "After" are the opposite—I get the warning when `proxy_hide_header X-Frame-Options;` is specified, otherwise I don't. My Nextcloud container's `site-confs/default.conf` has ([as per default](https://github.com/linuxserver/docker-nextcloud/blob/11660059f65370476dd05b2c9224ceef79eb398a/root/defaults/nginx/site-confs/default.conf.sample#L52))
```nginx
add_header X-Frame-Options "SAMEORIGIN" always;
```
Is this not the expected behavior?

nemchik:
Ok, I did a little more testing, and here is what I have found.

We ship https://github.com/linuxserver/docker-baseimage-alpine-nginx/blob/1cf9b27a72ac8e0787b577e9bc88767723b8615d/root/defaults/nginx/ssl.conf.sample#L37 commented out, and if you leave it commented out then nextcloud works because it is handling that header. The issue arises when you want other things being proxied by swag to also have that header included, in which case you would uncomment that line, which causes nextcloud to NOT work correctly unless you add `proxy_hide_header X-Frame-Options;` to the nextcloud proxy conf (as this PR does).

Documenting that in comments within the configs might be a bit confusing and verbose, so I'll have to think through it a bit, but for now the way to get it working would be to do one of the following:
- keep `add_header X-Frame-Options "SAMEORIGIN" always;` commented in `ssl.conf` and remove `proxy_hide_header X-Frame-Options;` from the nextcloud proxy
- uncomment `add_header X-Frame-Options "SAMEORIGIN" always;` in `ssl.conf` and include `proxy_hide_header X-Frame-Options;` in the nextcloud proxy

cdelledonne:
@nemchik thanks a lot for the very quick response! For now I will leave ssl.conf as per default (`#add_header X-Frame-Options "SAMEORIGIN" always;` commented) and remove `proxy_hide_header` from nextcloud.subdomain.conf.

> Weird because it's working for me with:
> uncommented `add_header X-Frame-Options "SAMEORIGIN" always` in swag-docker's ssl.conf
> and commented `#proxy_hide_header X-Frame-Options;` in nextcloud.subdomain.conf

@zedocrob that is actually not the case on my end. Perhaps it has to do with your Nextcloud image being too old and the site-confs/default.conf thereof not handling the header itself? The version of that file on my end is "Version 2022/08/20", which you should be able to obtain running this in your host (assuming you're using Docker and your Nextcloud container is named `nextcloud`):

```sh
docker exec -it nextcloud sed -r '1!d;s/## (Version [0-9/]+).+/\1/' /config/nginx/site-confs/default.conf
```

zedocrob:
I've used this guide to optimize Nextcloud
[https://virtualize.link/nextcloud/](url)
(now using `:latest` instead of` :php8` of course) i'm on latest version.
My ssl.conf is the latest:
[https://privatebin.net/?2cc71d7dfd2e302a#8V8b5eJSjoHMj9bHBbZXdjkvK9URAzbtAW5Kto7YPDbV](url)
default.conf is latest too:
[https://privatebin.net/?cf7559a8dfd0f394#5vP3qmP6AK26FiUcPo1zC3RzcKsnqSFyMwJLvhh7GBkP](url)
and nextcloud.subdomain.conf too Version 2022/10/28 with:
`#proxy_hide_header X-Frame-Options;`
 ` proxy_max_temp_file_size 2048m;`
  `add_header Strict-Transport-Security "max-age=15768000; includeSubDomains; preload";` (i've added this one as recommended so maybe not necessary?)

cdelledonne:
Perhaps I wasn't too clear, but I was looking for site-confs/default.conf in the Nextcloud container, not in the SWAG container

zedocrob:
I see! Sorry my mistake
sorry for noob question but where is default.conf located in the container?... :)
i've only found root@oiuyer789ljkk:/defaults/nginx/site-confs/default.conf.sample

cdelledonne:
It doesn't seem to be [the latest](https://github.com/linuxserver/docker-nextcloud/blob/11660059f65370476dd05b2c9224ceef79eb398a/root/defaults/nginx/site-confs/default.conf.sample), but I'm not sure if that's the reason for the behavior you're observing, perhaps you can try to pull the latest image and re-create the container?

zedocrob:
in `/defaults/nginx/site-confs/default.conf.sample` **inside the container** i have version 2022/08/20
https://privatebin.net/?b5be3f2f59904848#6PSAMZyaE6hkhSgd4PtmxiuwRbhj9av8HiFLBy3UC2hW
in `/config/nginx/site-confs/default.conf`  **inside the container** i have this:
https://privatebin.net/?8a973e7353f3ef55#Bbaa5ZPuBJ2wycrB4DVVRoGTS3kTSeRT4qvGFqj1WieP

`default.conf.sample` and `default.conf` looks different yes, is it normal?

nextcloud container version is :
Linuxserver.io version:- 25.0.1-ls211 Build-date:- 2022-11-13T09:37:13-06:00
I use Diun to update all my containers everyday so should be the latest.

cdelledonne:
`*.conf.sample` files are replaced by their latest versions upon update, but `*.conf` files are not. If you did not modify `default.conf` yourself, you can delete it and restart the container, and you'll get a fresh `default.conf` which will be the same as `default.conf.sample`. But again, I'm not sure this is the reason for the behavior you're observing.

zedocrob:
Thanks, good to know.
i did remove default.conf, recreate the container and went with this error:
`nginx: [emerg] no "ssl_certificate" is defined for the "listen ... ssl" directive in /config/nginx/site-confs/default.conf:9`
So i added to the new `default.conf`:
server_name _;
`ssl_certificate /config/keys/cert.crt;`
`ssl_certificate_key /config/keys/cert.key;`

Also i changed:
`client_max_body_size 512M;`
to `client_max_body_size 10G;` as my previous `defaul.conf`

So far so good i'm logged in without any error.
Is there another .conf file i should renew other than default.conf in container maybe?

cdelledonne:
There are `nginx.conf` and `ssl.conf`, you can do the same with those if you haven't modified them

PR #509, opened 2022-11-25
Create pterodactyl.conf.sample

vastofficial:


Roxedus (changes requested):
This is formated with too many spaces, aswell as not following the standard for declaring scheme, port and ip.

nemchik (changes requested):
This deviates significantly from our config templates. It should be based on 
and the file name should be pterpdactyl.subdomain.conf.sample

PR #510, opened 2022-12-01
Create wrapperr.subdomain.conf.sample

MountainGod2:

### DIFF
--- a/wrapperr.subdomain.conf.sample
+++ b/wrapperr.subdomain.conf.sample
@@ -1,0 +1,39 @@
+## Version 2022/09/08
+# make sure that your dns has a cname set for wrapperr and that your wrapperr container is not using a base url
+
+server {
+    listen 443 ssl;
+    listen [::]:443 ssl;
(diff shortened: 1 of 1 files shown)


MountainGod2:
Docker hub page to verify ports and name are correct:
https://hub.docker.com/r/aunefyren/wrapperr

nemchik (changes requested):

Ref:

MountainGod2:
Just curious, regarding the version date at the top of the file, is that the template date, or the date the config was created? I updated to today's date, so if that needs to be adjusted, no problem.

thespad:
> Just curious, regarding the version date at the top of the file, is that the template date, or the date the config was created? I updated to today's date, so if that needs to be adjusted, no problem.

It's the date the conf was created or last significantly modified, it's used to notify users when something important has changed and they may need to update their version of the file.

nemchik (approved):
lgtm, thanks!

PR #512, opened 2022-12-26
Instructions for setting Environment variable

coreyaboy:
Line 2 of this file instructs the user to set an environment variable in their filebrowser container, but supplies an incorrect name for that variable.

Filebrowser expects environment variables to be prepended with FB_ as per https://filebrowser.org/cli/filebrowser#options and as mentioned by one of the filebrowser contributors over at https://github.com/filebrowser/filebrowser/issues/1557#issuecomment-921756628


aptalca (dismissed):
Don't forget to update the version date
Thanks

nemchik (approved):

PR #549, opened 2023-04-10
Kasmvnc changes

thelamer:
This updates the examples for boinc and calibre for the KasmVNC base images to at least function with a subfolder configuration. 

There will be collision if multiple containers with the same KasmVNC base are used, but they can be modified to different paths in the client itself if the user wants to enable multiples, client settings: 

subfolder

I figured the best support would be to provide a working configuration and deal with


thelamer:
Linking https://github.com/linuxserver/docker-calibre/issues/129

aptalca (approved):

PR #550, opened 2023-04-13
Additional cadvisor endpoints

nemchik:

 - [ ] I have read the [contributing](https://github.com/linuxserver/reverse-proxy-confs/blob/master/.github/CONTRIBUTING.md) guideline and understand that I have made the correct modifications

### DIFF
--- a/cadvisor.subdomain.conf.sample
+++ b/cadvisor.subdomain.conf.sample
@@ -53,22 +53,4 @@ server {
         proxy_pass $upstream_proto://$upstream_app:$upstream_port;
 
     }
-
-    location ~ (/cadvisor)?/metrics {
-        # enable the next two lines for http auth
-        #auth_basic "Restricted";
-        #auth_basic_user_file /config/nginx/.htpasswd;
-
-        # block metrics access by default because it is unprotected
-        # you can comment out the next line to enable remote metrics
-        deny all;
-
-        include /config/nginx/proxy.conf;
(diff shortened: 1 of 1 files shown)


cadvisor.subdomain.conf.sample
```diff
@@ -41,5 +41,34 @@ server {
...
+    }
+
+    location ~ (/cadvisor)?/metrics {
+        # enable the next two lines for http auth
+        #auth_basic "Restricted";
+        #auth_basic_user_file /config/nginx/.htpasswd;
+
+        # block metrics access by default because it is unprotected
+        # you can comment out the next line to enable remote metrics
+        deny all;
```
aptalca, line 64: This behavior would break the auto-proxy mod. Can we assume that users shouldn't publicly expose cadvisor without auth to begin with?

Normally our thinking process is, assume everything will be put behind auth (we provide 4 different methods that are built in), then break out things where the auth should be bypassed (like api). We don't account for things that should be added to auth, or specifically blocked because we assume everything will be behind auth.

nemchik: Removed metrics endpoint in https://github.com/linuxserver/reverse-proxy-confs/pull/550/commits/99490181017b4e61ad7e981918aae212bf678f4d
per discussion in discord; metrics endpoint can be reached via the root location. I will open a separate PR to remove the metrics endpoints with `deny`.

aptalca (approved):

PR #556, opened 2023-04-16
check for nginx conf validity

aptalca:
templates are not valid and are excluded
organizr, heimdall and wordpress subfolder confs duplicate `location /` and thus excluded
boinc subfolder duplicates `location /websockify` with calibre and thus excluded

All other subdomain and subfolder proxy conf samples are added to nginx config on the fly and checked for validity

Successful PR result: https://github.com/linuxserver/reverse-proxy-confs/actions/runs/4714993511
Broken PR result: https://github.com/linuxserver/reverse-proxy-confs/actions/runs/4715002705


nemchik (approved):
lgtm

PR #560, opened 2023-04-18
add kasm subdomain sample

thelamer:
Adds a subdomain conf for both the main Kasm interface and the wizard with notes about post install configuration.

### DIFF
--- a/kasm.subdomain.conf.sample
+++ b/kasm.subdomain.conf.sample
@@ -1,5 +1,6 @@
 ## Version 2023/04/18
 # make sure that your kasm container is named kasm
+# make sure that your dns has a cname set for kasm and kasm-wizard
 
 # This configuration assumes 8443 with the environment variable -e KASM_PORT=8443 set adjust to your needs
 # Post installation you will need to access Kasm > Admin > Zones > default zone (edit) and modify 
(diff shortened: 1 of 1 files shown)


kasm.subdomain.conf.sample
```diff
@@ -0,0 +1,83 @@
...
+    client_max_body_size 0;
+
+    location / {
+        proxy_set_header        Upgrade $http_upgrade;
+        proxy_set_header        Connection "upgrade";
+        proxy_set_header        Host $host;
+        proxy_set_header        X-Real-IP $remote_addr;
+        proxy_set_header        X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_set_header        X-Forwarded-Proto $scheme;
+        proxy_http_version      1.1;
```
Roxedus, line 24: Any reason you are shipping these, rather than including, and inheriting from proxy conf?

thelamer: I thought it was safest to follow the Kasm example to the T:  https://www.kasmweb.com/docs/latest/how_to/reverse_proxy.html#example-nginx-config

But you do make a valid point about many of them being identical, was concerned about additional vars messing with it, I will test on the VDI server and see if it has any effect.

thelamer: I remember why now it was the timeout values contradicting with the base, but I think those are very excessive and 240 is fine. I just tested on the VDI server in germany and it still worked great, it wont get much higher latency for anyone else.

nemchik: Linking for quick reference https://github.com/linuxserver/docker-swag/blob/master/root/defaults/nginx/proxy.conf.sample

The default from nginx for `proxy_connect_timeout` is `60`, and it looks like our value of `240` dates all the way back to the original addition of the proxy file in the former le repo https://github.com/linuxserver/docker-letsencrypt/commit/7a5926c8322b37330bd15ef59794c1d13b40863a

I'm not opposed to revisiting the values we ship in the proxy conf. I know most of the more recent changes and additions are as close to stock or nginx recommendations as possible, sprinkled in are things required for authelia/authentik.

nemchik:
Would there be any benefit to shipping this as two conf files (one for each server block)? Do these two things ever run as two separate containers?

kasm.subdomain.conf.sample
thelamer: I think the 1800 in the Kasm example was likely just a "max it out" style decision to deal with problem deployments, the 240 works fine for me and worked in my testing.

thelamer:
> Would there be any benefit to shipping this as two conf files (one for each server block)? Do these two things ever run as two separate containers?

Naw it's always the same container, this container is weird that it has a web interface to properly deploy the other web interface. It is also used for upgrades. It was a way to deploy the Kasm app in an isolated DinD container instead of requiring host level installs that are not gonna be compatible with most home server solutions.

nemchik (changes requested):

nemchik (approved):

PR #574, opened 2023-05-27
fix: bypass auth for kavita api endpoint

YodaDaCoda:

## Description

Bypass authelia etc for api access to kavita - permits use of OPDS endpoints while auth is in use.

With the current config, when using one of the auth options, access to the OPDS endpoint is captured by nginx & redirected to auth. This adds a bypass for the /api endpoint which uses a token system. 

## How Has This Been Tested?

Implemented the changes on my local system & tested access with Librera on android. Does not work without this change, works with.

## Source / References

Changes taken from Kavita's docs: https://wiki.kavitareader.com/en/install/access-kavita-from-network/reverse-proxy/swag-example

### DIFF
--- a/kavita.subdomain.conf.sample
+++ b/kavita.subdomain.conf.sample
@@ -43,4 +43,15 @@ server {
         proxy_pass $upstream_proto://$upstream_app:$upstream_port;
 
     }
+
+    # Needed for OPDS access while using Authelia/ldap
+    location ~ /api {
+        include /config/nginx/proxy.conf;
+        include /config/nginx/resolver.conf;
+        set $upstream_app kavita;
+        set $upstream_port 5000;
+        set $upstream_proto http;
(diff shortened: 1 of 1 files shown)


kavita.subdomain.conf.sample
```diff
@@ -43,4 +43,15 @@ server {
         proxy_pass $upstream_proto://$upstream_app:$upstream_port;
 
     }
+
+    # Needed for OPDS access while using Authelia/ldap
+    location ~ /api {
```
nemchik, line 48:
```suggestion
    location ~ (/kavita)?/api {
```
For the sake of consistency with other confs, please include this (even though it's likely not used in any circumstance).

nemchik (changes requested):

nemchik (approved):

PR #578, opened 2023-06-06
Turn off proxy_buffering for nextcloud

nemchik:


thespad:
Is this just a QOL thing or is there a direct impact to how it behaves?

nemchik:
> Is this just a QOL thing or is there a direct impact to how it behaves?

It could have a direct (positive) impact for users uploading large files.
https://docs.nextcloud.com/server/latest/admin_manual/configuration_files/big_file_upload_configuration.html#nginx
The last couple bullet points under the nginx section say to either increase `proxy_max_temp_file_size` (which we were, but that's being replaced) or disable `proxy_buffering` (which this does). We don't need to do both, because disabling it means whatever size we set isn't used, so replacing made more sense than appending. The reason I'm opting for disabled by default (rather than increase) is nextcloud docs don't have a specific size recommendation for this setting. They recommend 16gb for some PHP settings, so they aren't shy with big numbers, but they didn't specify this one.

thespad (approved):

nemchik:
> Is there a similar change going into the Nextcloud default conf?

There are changes coming to nextcloud, but not related to this change. This applies specifically to the proxy.

PR #582, opened 2023-06-20
Fix organizr settings to fit v2.1

ofersadan85:

## Description

Organizr updated to version 2.1 and the previous conf file stopped working properly. The new conf needs a location for `/api/v2` as well, see [here](https://docs.organizr.app/help/faq/migration-guide#version-2-0-greater-than-version-2-1)

Organizr will no longer work with the previous proxy settings

## How Has This Been Tested?
Tested work with the default [organizr/organizr](https://hub.docker.com/r/organizr/organizr) image

## Source / References
https://docs.organizr.app/help/faq/migration-guide#version-2-0-greater-than-version-2-1

### DIFF
--- a/organizr.subdomain.conf.sample
+++ b/organizr.subdomain.conf.sample
@@ -1,4 +1,4 @@
-## Version 2023/05/31
+## Version 2023/06/20
 # make sure that your organizr container is named organizr
 # make sure that your dns has a cname set for organizr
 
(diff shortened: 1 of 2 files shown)


Roxedus:
This should not be needed. The organizr image already handles this endpoint. 

> If you are using or you have now switched over to this container, you will not need to add the new api location block as that has been added for you in the container image.

-The linked documentation 

I do not currently, nor have I ever needed to have the endpoint in my proxy-conf for my Organizr reverse proxy.

(It seems like you do not have this endpoint in the nginx config for the organizr container, which you should troubleshoot elsewhere)

PR #600, opened 2023-08-28
Add `/s` auth bypass for Photoprism

ajnart:
`/s` is the path that corresponds to semi-publicly available share URLs in Photoprism (they still require a secret key in the URL), this commit make it so that they are bypassed by the authentification method.

As referenced in https://github.com/linuxserver/docker-mods/issues/372#issuecomment-1696145759


ajnart:
As mentioned in the issue above, he client requests the url in /library/albums/ when using a /s/ link. Therefore it's redirected to authelia and forced to log-in, making these changes useless.

PR #607, opened 2023-09-05
Update komga.subdomain.conf.sample - updated the port to 25600 from 8080

realbuz:
2023/09/05

docker has recently updated the port to 25600 from 8080

allows new deployment to work without end user having to modify the sample file manually

## How Has This Been Tested?
made the change on the file manually in my swag deployment without issues

## Source / References
in the breaking changes: https://komga.org/blog/prepare-v1/

### DIFF
--- a/komga.subdomain.conf.sample
+++ b/komga.subdomain.conf.sample
@@ -1,4 +1,4 @@
-## Version 2023/05/31
+## Version 2023/09/05
 # make sure that your komga container is named komga
 # make sure that your dns has a cname set for komga
 
--- a/komga.subdomain.conf.sample
+++ b/komga.subdomain.conf.sample
@@ -38,7 +38,7 @@ server {
         include /config/nginx/proxy.conf;
         include /config/nginx/resolver.conf;
         set $upstream_app komga;
-        set $upstream_port 8080;
+        set $upstream_port 25600;
         set $upstream_proto http;
         proxy_pass $upstream_proto://$upstream_app:$upstream_port;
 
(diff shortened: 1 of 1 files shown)


aptalca (changes requested):
Thanks for the PR, but please update the version date

drizuid:
> anyone using docker and keeping their images somewhat up to date will be on the new port.

did the dev end of life/end of support the old version?

realbuz:
> > anyone using docker and keeping their images somewhat up to date will be on the new port.
> 
> did the dev end of life/end of support the old version (pre v1.0.0)?

dockerhub goes as far back as the 0.1.0 version 4 years ago.  Github goes back 12 months ago in the tagged version 0.157.2.

not sure if there's an actual lifecycle process to have version end of like/support. 

My view is that the breaking changes were announced prior to version 1 being released by the komga dev team in advance, I'm of the opinion that swag should follow the same.

Most people who have been using swag will know how to fix the port after updating komga. Anyone who's deploying a net new swag/docker environment shouldn't have to reconfigure things to get going.  again, just my opinion.

drizuid:
@aptalca i'll leave it up to you, i'm not sure if we have precedent for this situation or not

nemchik (approved):
assuming the current version uses this port, existing users who have not updated komga won't be affected by this change unless they take action to update the config. swag will notify them of an out of date config.

auto proxy mod users may be a different story.

regardless, if a user installed komga today they would likely not pick an old version using port 8080, so they would need to adjust the port in the proxy conf.

PR #633, opened 2023-12-14
Secure cookies in heimdall.subdomain.conf.sample

d4rklynk:

## Description
The cookies are not secure by default with heimdall, we can force it through nginx.

## How Has This Been Tested?
The cookies are functional and can be seen through developer console on my instance -> https://services.simpleprivacy.fr/

image

### DIFF
--- a/heimdall.subdomain.conf.sample
+++ b/heimdall.subdomain.conf.sample
@@ -34,7 +34,7 @@ server {
 
         # enable for Authentik (requires authentik-server.conf in the server block)
         #include /config/nginx/authentik-location.conf;
-        
+
         include /config/nginx/proxy.conf;
         include /config/nginx/resolver.conf;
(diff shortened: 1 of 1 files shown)


nemchik (commented):
We have a similar line here https://github.com/linuxserver/docker-swag/blob/5b77a546209cab487853d3b0d3505578d11e7f52/root/defaults/nginx/proxy.conf.sample#L18 with a warning about applying it. There are some applications that have issues with this configuration, so it's not likely appropriate to be used globally (in the `proxy.conf`) but may be safe in an individual proxy conf for an application (like Heimdall) as you have it here.

I would prefer it to be commented out by default, with a comment description, and placed below the `proxy_pass $upstream_proto://$upstream_app:$upstream_port;` line

nemchik (approved):

PR #644, opened 2024-01-17
grpc section

drizuid:
add grpc support for the app

### DIFF
--- a/pydio-cells.subdomain.conf.sample
+++ b/pydio-cells.subdomain.conf.sample
@@ -1,4 +1,4 @@
-## Version 2023/05/31
+## Version 2024/01/17
 # make sure that your pydio-cells container is named pydio-cells
 # make sure that your dns has a cname set for pydio-cells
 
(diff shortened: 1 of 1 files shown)


drizuid:
thanks to thedinz

pydio-cells.subdomain.conf.sample
```diff
@@ -41,6 +41,9 @@ server {
         set $upstream_port 8080;
         set $upstream_proto https;
         proxy_pass $upstream_proto://$upstream_app:$upstream_port;
+        if ($http_content_type = "application/grpc") {
+            grpc_pass grpcs://pydio-cells:33060;
```
Roxedus, line 45: Should be the $upstream_app:$upstream_port pattern.

Roxedus (changes requested):

drizuid:
putting this on hold as I don't use pydio-cells and the [documentation ](https://pydio.com/en/docs/kb/deployment/running-cells-behind-nginx-reverse-proxy)implies that the port used is the normal 8080 port. need to get confirmation/testing from someone that uses this.
image

pydio-cells.subdomain.conf.sample
aptalca: done

aptalca:
upstream now defaults to 8080 for grpc, var is removed from our pydio-cells image in https://github.com/linuxserver/docker-pydio-cells/pull/17

aptalca (approved):

Roxedus (approved):

PR #654, opened 2024-03-03
Created rallly.subdomain.conf.sample

rastacalavera:
subdomain proxy config for Rallly App

## Description

I recently discovered Rallly which is scheduling software similar to Doodle (polling with Calendar). I created a proxy config that I tested and can verify that it works so that if others want to use this software with SWAG, they can easily.

Allows the use of user funded software to be used with SWAG easily. 
## How Has This Been Tested?

I tested the subdomain only since that is the approach that I use. 
## Source / References

[Rallly](https://rallly.co)

### DIFF
--- a/rallly.subdomain.conf.sample
+++ b/rallly.subdomain.conf.sample
@@ -1,0 +1,46 @@
+## Version 2023/05/31
+# make sure that your <container_name> container is named rallly
+# make sure that your dns has a cname set for rallly
+
+server {
+    listen 443 ssl http2;
+    listen [::]:443 ssl http2;
+
+    server_name rallly.*;
+
+    include /config/nginx/ssl.conf;
+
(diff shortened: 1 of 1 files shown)


rallly.subdomain.conf.sample
```diff
@@ -0,0 +1,46 @@
+## Version 2023/05/31
+# make sure that your <container_name> container is named rallly
```
drizuid, line 2: <container_name> is a placeholder, in your case, it should be rally, please adjust for approval. thanks!

drizuid (changes requested):

rallly.subdomain.conf.sample
rastacalavera: change has been made, thanks!

drizuid (approved):

PR #655, opened 2024-03-08
add support for Nextcloud notify push mod

aptalca:
Location block needed by the new notify push mod:
https://github.com/linuxserver/docker-mods/pull/852


nemchik (approved):
approving, pending review on https://github.com/linuxserver/docker-mods/pull/852 which needs readme adjustments to properly document the changes here

aptalca:
Decided to add it to NC's site conf instead: https://github.com/linuxserver/docker-nextcloud/pull/423
That way it will work with all reverse proxies out of the box

PR #659, opened 2024-03-22
Create kimai.subdomain.conf.sample

drizuid:
did not test API functionality as I'm not using it

 - [ ] I have read the [contributing](https://github.com/linuxserver/reverse-proxy-confs/blob/master/.github/CONTRIBUTING.md) guideline and understand that I have made the correct modifications

### DIFF
--- a/kimai.subdomain.conf.sample
+++ b/kimai.subdomain.conf.sample
@@ -45,19 +45,6 @@ server {
     }
 
     location ~ (/kimai)?/api {
-        # enable the next two lines for http auth
-        #auth_basic "Restricted";
-        #auth_basic_user_file /config/nginx/.htpasswd;
-
-        # enable for ldap auth (requires ldap-server.conf in the server block)
-        #include /config/nginx/ldap-location.conf;
-
-        # enable for Authelia (requires authelia-server.conf in the server block)
-        #include /config/nginx/authelia-location.conf;
-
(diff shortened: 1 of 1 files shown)


kimai.subdomain.conf.sample
```diff
@@ -44,4 +44,27 @@ server {
...
+
+        # enable for ldap auth (requires ldap-server.conf in the server block)
+        #include /config/nginx/ldap-location.conf;
+
+        # enable for Authelia (requires authelia-server.conf in the server block)
+        #include /config/nginx/authelia-location.conf;
+
+        # enable for Authentik (requires authentik-server.conf in the server block)
+        #include /config/nginx/authentik-location.conf;
+
```
nemchik, line 60: we don't usually put the auth lines here
https://github.com/linuxserver/reverse-proxy-confs/blob/6b2e31a42bb6eceb58bc663251861b4a691c7b19/_template.subdomain.conf.sample#L51-L62
not that they can't be, but if we're going to include them here it would be better to make a separate PR and include them in all api blocks for all confs.

nemchik (changes requested):

kimai.subdomain.conf.sample
drizuid: my bad on the delay, forgot i submitted this :D

nemchik (approved):

PR #662, opened 2024-03-28
Privatebin needs to use http for upstream

aptalca:
https://github.com/PrivateBin/docker-nginx-fpm-alpine?tab=readme-ov-file#behind-the-scenes


quietsy (approved):
LGTM

PR #666, opened 2024-03-31
Add configurations for Sonarr to RSS

gbendy:
Add proxy configurations for Sonarr to RSS

 - [ X ] I have read the [contributing](https://github.com/linuxserver/reverse-proxy-confs/blob/master/.github/CONTRIBUTING.md) guideline and understand that I have made the correct modifications

## Description

Sonarr to RSS is a new Sonarr helper app.  

https://github.com/gbendy/sonarrToRSS/
https://registry.hub.docker.com/r/gbendy/sonarrtorss/

This PR adds both subdomain and subfolder reverse proxy support to Swag. The application supports both types of reverse proxy natively.

Users can configure swag without having to manually install configurations files from the app repository

## How Has This Been Tested?

Have been running the subdomain configuration myself for over a week with no problems.

Have run subfolder proxy for long enough to test that all endpoints work.

Have run both with authentication enabled and ensured that the 'api' endpoints are exempt from authentication and the user endpoints are not.

## Source / References

N/A

### DIFF
--- a/sonarrtorss.subdomain.conf.sample
+++ b/sonarrtorss.subdomain.conf.sample
@@ -1,0 +1,54 @@
+## Version 2024/06/21
+# make sure that your sonarrtorss container is named sonarrtorss
+# make sure that your dns has a cname set for sonarrtorss
+
+server {
+    listen 443 ssl http2;
+    listen [::]:443 ssl http2;
+
+    server_name sonarrtorss.*;
+
+    include /config/nginx/ssl.conf;
+
+    client_max_body_size 0;
+
+    # enable for ldap auth (requires ldap-location.conf in the location block)
+    #include /config/nginx/ldap-server.conf;
+
+    # enable for Authelia (requires authelia-location.conf in the location block)
+    #include /config/nginx/authelia-server.conf;
+
+    # enable for Authentik (requires authentik-location.conf in the location block)
+    #include /config/nginx/authentik-server.conf;
+
+    location / {
+        # enable the next two lines for http auth
+        #auth_basic "Restricted";
+        #auth_basic_user_file /config/nginx/.htpasswd;
+
+        # enable for ldap auth (requires ldap-server.conf in the server block)
+        #include /config/nginx/ldap-location.conf;
+
+        # enable for Authelia (requires authelia-server.conf in the server block)
+        #include /config/nginx/authelia-location.conf;
+
+        # enable for Authentik (requires authentik-server.conf in the server block)
+        #include /config/nginx/authentik-location.conf;
+
+        include /config/nginx/proxy.conf;
+        include /config/nginx/resolver.conf;
+        set $upstream_app sonarrtorss;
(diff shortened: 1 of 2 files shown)


drizuid (approved):

sonarrtorss.subdomain.conf.sample
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    location ~ /json {
+        include /config/nginx/proxy.conf;
+        include /config/nginx/resolver.conf;
+        set $upstream_app sonarrtorss;
+        set $upstream_port 18989;
+        set $upstream_proto http;
+        proxy_pass $upstream_proto://$upstream_app:$upstream_port;
+    }
```
Roxedus, line 89: Whats the point of these locations? they should be all covered by /

sonarrtorss.subfolder.conf.sample
```diff
@@ -0,0 +1,82 @@
...
+
+location ^~ /sonarrtorss/json {
+    include /config/nginx/proxy.conf;
+    include /config/nginx/resolver.conf;
+    set $upstream_app sonarrtorss;
+    set $upstream_port 18989;
+    set $upstream_proto http;
+    proxy_pass $upstream_proto://$upstream_app:$upstream_port;
+
+    rewrite /sonarrtorss(.*) $1 break;
```
Roxedus, line 81: Samme applies to these

Roxedus (changes requested):

sonarrtorss.subfolder.conf.sample
gbendy: Because these endpoints are ones that you would typically not want the authentication configured in the `/` to be applied to. Feed endpoints are usually unauthenticated and the `/sonarr` needs to use the authentication configured within the application

Roxedus: a solution like this is more fitting then https://github.com/linuxserver/reverse-proxy-confs/blob/0947948124e5a9f5d085b13a79a38c212694546f/homeassistant.subdomain.conf.sample#L56

sonarrtorss.subfolder.conf.sample
```diff
@@ -0,0 +1,36 @@
...
+}
+
+location ~ ^/(api/|sonarr$|rss$|atom$|json$) {
```
Roxedus, line 29: This need the subfolder prefixed, as this will break other apps using ie api in their path.
https://github.com/linuxserver/reverse-proxy-confs/blob/master/sonarr.subfolder.conf.sample#L28

Roxedus (changes requested):

sonarrtorss.subfolder.conf.sample
gbendy: Whoops, so it did. And it needed the rewrite.

Roxedus (approved):

PR #669, opened 2024-04-13
Created openproject.conf.sample

HammyHavoc:
This is for https://www.openproject.org/

### DIFF
--- a/openproject.conf.sample
+++ b/openproject.conf.sample
@@ -1,3 +1,7 @@
+## Version 2024/04/13
+# make sure that your OpenProject container is named OpenProject
+# make sure that your DNS has a record set for openproject
+
 server {
     listen 443 ssl;
     listen [::]:443 ssl;
--- a/openproject.conf.sample
+++ b/openproject.conf.sample
@@ -1,0 +1,18 @@
+server {
+    listen 443 ssl;
+    listen [::]:443 ssl;
+
+    server_name openproject.*;
+    include /config/nginx/ssl.conf;
+    client_max_body_size 0;
+
+    location / {    
+        include /config/nginx/proxy.conf;
+        include /config/nginx/resolver.conf;
+        set $upstream_app OpenProject;
+        set $upstream_port 8080;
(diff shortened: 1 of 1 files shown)


openproject.conf.sample
```diff
@@ -0,0 +1,22 @@
+## Version 2024/04/13
+# make sure that your OpenProject container is named OpenProject
```
drizuid, line 2: I believe there are issues on some systems with upper case characters

openproject.conf.sample
```diff
@@ -0,0 +1,22 @@
...
+    listen [::]:443 ssl;
+
+    server_name openproject.*;
+    include /config/nginx/ssl.conf;
+    client_max_body_size 0;
+
+    location / {    
+        include /config/nginx/proxy.conf;
+        include /config/nginx/resolver.conf;
+        set $upstream_app OpenProject;
```
drizuid, line 16: I believe there are issues on some systems with upper case characters

drizuid (commented):
@nemchik can you confirm?

nemchik:
> @nemchik can you confirm?

Confirmed. Some systems force lowercase container name resolution.

openproject.conf.sample
```diff
@@ -0,0 +1,22 @@
...
+        set $upstream_port 8080;
+        set $upstream_proto https;
+        proxy_pass $upstream_proto://$upstream_app:$upstream_port;
+        proxy_set_header X-Forwarded-Host $upstream_app:$upstream_port;
```
nemchik, line 20: This seems wrong. We have `X-Forwarded-Host` set here
https://github.com/linuxserver/docker-swag/blob/41cb7cb10423ac9b2708a55b94e6901955199d49/root/defaults/nginx/proxy.conf.sample#L28
We intentionally use the `$host` variable as it provides information about the visiting client to the header, and thus to the application.

The values you are passing here would essentially be static, and not reflect anything about the client. Is there a documented requirement for this?

PR #684, opened 2024-07-06
Update nextcloud's instructions

quietsy:
We're getting a lot of support questions regarding `trusted_proxies` and `gethostbyname`, changing the instructions to be more fool-proof

### DIFF
--- a/nextcloud.subdomain.conf.sample
+++ b/nextcloud.subdomain.conf.sample
@@ -1,18 +1,15 @@
-## Version 2024/04/25
+## Version 2024/07/06
 # make sure that your nextcloud container is named nextcloud
+# make sure that your nextcloud container is on the same docker network as swag
 # make sure that your dns has a cname set for nextcloud
-# assuming this container is called "swag", edit your nextcloud container's config
-# located at /config/www/nextcloud/config/config.php and add the following lines before the ");":
-#  'trusted_proxies' => [gethostbyname('swag')],
+# edit your nextcloud container's config located at
+# /config/www/nextcloud/config/config.php and add the following lines before the ");":
+#  'trusted_proxies' => ['172.16.0.0/12'],
 #  'overwrite.cli.url' => 'https://nextcloud.example.com/',
 #  'overwritehost' => 'nextcloud.example.com',
 #  'overwriteprotocol' => 'https',
-#
-# Also don't forget to add your domain name to the trusted domains array. It should look somewhat like this:
-#  array (
-#    0 => '192.168.0.1:444', # This line may look different on your setup, don't modify it.
-#    1 => 'nextcloud.example.com',
-#  ),
+# add your domain name to the trusted domains array:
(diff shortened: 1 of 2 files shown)


thespad:
Need to be careful as you're making assumptions about `172.16.0.0/12`. Docker can also allocate /20s from `192.168.0.0/16` and, in the case of overlay networks, /24s from `10.0.0.0/8`.

Technically it can allocate any sized subnet from those ranges, but those are the defaults and you'd hope if people have the wherewithal to change them they also know how to configure this without handholding.

quietsy:
> Need to be careful as you're making assumptions about `172.16.0.0/12`. Docker can also allocate /20s from `192.168.0.0/16` and, in the case of overlay networks, /24s from `10.0.0.0/8`.
> 
> Technically it can allocate any sized subnet from those ranges, but those are the defaults and you'd hope if people have the wherewithal to change them they also know how to configure this without handholding.

It will work in most cases and specifically in the common compose created network, and avoid the issues with `gethostbyname`.

nemchik:
I searched through discord briefly and the only real issues I found with `gethostbyname` were that at some point NC (or a plugin) replaced it in the `config.php` with an actual IP address for SWAG at the time (which could later change). I can confirm this actually happened to me. The rest of what I saw regarding `trusted_domains` was mostly users who didn't have `gethostbyname` involved, or didn't have their NC and their proxy container (swag or otherwise) on the same docker network.

I had put `gethostbyname` in my config a few months ago and found today it was replaced with `172.16.0.8` at some point. I think it makes the most sense to try our best to explain what's actually going on in the comments in the file so users can decide what value they want to put in their `config.php`, potentially from multiple known working options. I would consider `gethostbyname` a known working option with a caveat (that it will likely get replaced by a static IP and not work long term). Aside from that, using the CIDR for the docker network if the user has their containers on the same docker network, or using the CIDR for their LAN as a last resort.

quietsy:
I don't see a need to over-complicate it beyond: `gethostbyname` is known to break, using the CIDR doesn't break.

> if the user has their containers on the same docker network
It shouldn't be an if, that is our recommended configuration with swag and what officially we support.

nemchik:
> I don't see a need to over-complicate it beyond: `gethostbyname` is known to break, using the CIDR doesn't break.
> 
> > if the user has their containers on the same docker network
> 
> It shouldn't be an if, that is our recommended configuration with swag and what we officially support.

I can agree with both of those statements. Would it be worth writing a PHP script that detects the user's CIDR from within the container and displays it as a recommended value in the init log? Potentially even as an occ command the user can just copy/paste to set the correct value?

P.s. I don't think setting it automatically is the right idea, but displaying a recommendation for settings we support based on detected values in the user environment makes sense.

Also could be done in bash, but I figure if PHP can see the value it's more likely what nextcloud would actually see (depending on their implementation I guess).

spider1163:
please not stale

PR #692, opened 2024-08-17
include auth config files on Nextcloud

RMI78:
# Added includes for authentication mechanism for nextcloud

## Description

I simply added the different includes for Authelia, Authentik, ldap and http as a comment in the Nextcloud reverse proxy config file as those lines are also in other config files. 

This PR allow redirecting any request done to a Nextcloud instance behind SWAG to the authentication mechanism (if it exist) whereas it has to be entered manually before. As it is only commented code, nothing should be impacted and uncommenting those lines can be done at the user's discretion.

I was using SWAG with the docker auto-proxy-mod (and all it requirements), used Authelia with a wildcard to protect all my services behind (Nextcloud + SearXNG) even though it means logging twice (Authelia + Nextcloud), I don't mind.  The docker auto-proxy-mod automatically grab the config samples when the software is detected and has a generic template for others (including the lines I added here). It then proceeds to uncomment lines of any copied config files for any auth mechanism in function of their associated container label. In the case of Nextcloud, there was nothing to uncomment, thus, protecting my SearXNG instance and exposing my Nextcloud instance.

## How Has This Been Tested?

I just `docker exec -it bash swag` into my swag container, went to the /etc/nginx/http.d/auto-proxy-nextcloud.conf to add the lines for Authelia and I reloaded only the nginx server within the container with `/usr/sbin/nginx -c /config/nginx/nginx.conf -s reload`. Works like a charm now, but I would like to not do that every time neither configure it manually, so maybe for consistency of all the other config files here and for the sake of automation it could be a great idea to include those comments.

## Source / References

[Docker auto proxy bash file](https://github.com/linuxserver/docker-mods/blob/swag-auto-proxy/root/app/auto-proxy.sh) in case you need it.

### DIFF
--- a/nextcloud.subdomain.conf.sample
+++ b/nextcloud.subdomain.conf.sample
@@ -24,7 +24,29 @@ server {
 
     client_max_body_size 0;
 
+    # enable for ldap auth (requires ldap-location.conf in the location block)
+    #include /config/nginx/ldap-server.conf;
+
+    # enable for Authelia (requires authelia-location.conf in the location block)
+    #include /config/nginx/authelia-server.conf;
+
+    # enable for Authentik (requires authentik-location.conf in the location block)
+    #include /config/nginx/authentik-server.conf;
+
     location / {
+        # enable the next two lines for http auth
+        #auth_basic "Restricted";
+        #auth_basic_user_file /config/nginx/.htpasswd;
+
+        # enable for ldap auth (requires ldap-server.conf in the server block)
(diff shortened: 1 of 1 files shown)


aptalca:
Nextcloud mobile app and various addons break when it's behind another auth. Nextcloud is meant to be exposed directly and it contains various security functionality itself including 2fa and brute force protection. That's why we never added the auth bits.

RMI78:
alright alright, closing it then

PR #693, opened 2024-08-18
restrict vaultwarden admin page to LAN

jlssmt:

## Description
vaultwarden admin page should be restricted to local only

## Source / References
https://github.com/dani-garcia/vaultwarden/discussions/2626


jlssmt:
would prefer a internal.conf like described here:
https://virtualize.link/secure/

jlssmt:
would also prefer enabled by default

Roxedus:
I think this should mention why restricting /admin is a good idea

Roxedus (approved):

PR #705, opened 2024-10-17
change app name deets

drizuid:
I missed this but this project was renamed https://github.com/zwave-js/zwave-js-ui/releases/tag/v8.0.0 and I've updated the conf accordingly to avoid confusion, especially considering how long ago it was changed.


quietsy (approved):
LGTM

PR #727, opened 2024-12-30
Create headscale.subdomain.conf.sample

npcraft:

## Description

Headscale Nginx configs added. 

Headscale Nginx configs were missing, so it is added. It will help everyone who wants to configure headscale.

## How Has This Been Tested?

I have setup Headscale with tailscale client on phone and laptop to make sure that config works.

## Source / References

https://headscale.net/stable/ref/integration/reverse-proxy/?h=nginx

### DIFF
--- a/headscale.subdomain.conf.sample
+++ b/headscale.subdomain.conf.sample
@@ -1,0 +1,42 @@
+## Version 2024/12/29
+# make sure that your headscale container is named headscale
+# make sure that your dns has a cname set for headscale
+
+server {
+    listen 443 ssl;
+    listen [::]:443 ssl;
+
+    server_name headscale.*;
+
+    include /config/nginx/ssl.conf;
+
+    client_max_body_size 0;
+
+    # enable for ldap auth (requires ldap-location.conf in the location block)
+    #include /config/nginx/ldap-server.conf;
+
+    # enable for Authelia (requires authelia-location.conf in the location block)
+    #include /config/nginx/authelia-server.conf;
+
+    # enable for Authentik (requires authentik-location.conf in the location block)
+    #include /config/nginx/authentik-server.conf;
+
+    location / {
+        # enable for ldap auth (requires ldap-server.conf in the server block)
+        #include /config/nginx/ldap-location.conf;
+
+        # enable for Authelia (requires authelia-server.conf in the server block)
+        #include /config/nginx/authelia-location.conf;
+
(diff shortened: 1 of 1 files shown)


headscale.subdomain.conf.sample
```diff
@@ -0,0 +1,42 @@
...
+    # enable for ldap auth (requires ldap-location.conf in the location block)
+    #include /config/nginx/ldap-server.conf;
+
+    # enable for Authelia (requires authelia-location.conf in the location block)
+    #include /config/nginx/authelia-server.conf;
+
+    # enable for Authentik (requires authentik-location.conf in the location block)
+    #include /config/nginx/authentik-server.conf;
+
+    location / {
```
quietsy, line 24: Please add an `/api` location endpoint without the auth comments, you can take a look at other samples as a reference.

npcraft: @quietsy Actually /api location is not needed for this proxy to work. I have it in production already for long time, works perfectly fine.  
Please let me know, what do you think.

quietsy: It's not needed but it's our standard way of writing confs to avoid api issues when enabling authentik/authelia.

npcraft: @quietsy `/api` location is not added because in example file `_template.subdomain.conf.sample` mentioned that, if `/api` is not required for headscale it should not be included.
 
As this comment in the file confusing for me, can you please explain, as you have said it should be included. 

Otherpoint  that headscale is really not compatible with authentik and authelia, so I would like to also remove that commented lines, because it will be confusing to keep something which will not be compatible with headscale. 

Please let me know what do you think? Thanks for you time.

```
 # REMOVE THIS LINE BEFORE SUBMITTING: Some proxies require one or more additional location blocks for things like API or RPC endpoints.
# REMOVE THIS LINE BEFORE SUBMITTING: If the proxy you are making a sample for does not require an additional location block please remove the commented out section below.
    # location ~ (/<container_name>)?/api {
    #     include /config/nginx/proxy.conf;
    #     include /config/nginx/resolver.conf;
    #     set $upstream_app <container_name>;
    #     set $upstream_port <port_number>;
    #     set $upstream_proto <http or https>;
    #     proxy_pass $upstream_proto://$upstream_app:$upstream_port;
    #
    #     # REMOVE THIS LINE BEFORE SUBMITTING: Additional proxy settings such as headers go below this line, leave the blank line above.
    # }  
```

PR #728, opened 2025-01-01
for subfolder too... i guess they exist

drizuid:
large comics weren't able to sync to kobo with the settings due to the proxy_buffers setting in proxy.conf, remove the inclusion and provide the specific requirements in the location block itself, per upstream dev's docs


drizuid:
Tested and confirmed working by Mr grimace https://discord.com/channels/354974912613449730/506828136869265408/1324126662794940478

nemchik (approved):

PR #729, opened 2025-01-07
added hoarder

jlssmt:


drizuid:
were both of these tested? do they work fully? does api work? literally no info here.

jlssmt:
Yep. I'm using it and api is also working.

drizuid:
> Yep. I'm using it and api is also working.

in the future, please note when you have tested your PR and all functions of it. Can you please confirm you actually tested subfolder as well, since you state "im using it" rather than "im using them" and it'd be abnormal to use both subdomain and subfolder concurrently.

jlssmt:
you're right. i'm only using and have tested the subdomain config which is working.
the subfolder config is not working for me but maybe the problem is my environment.
i can remove the subfolder config from this PR.

for future i will only submit the subdomain config and write in the description that it's tested by me.

drizuid (approved):

PR #735, opened 2025-01-12
Create hoarder.subdomain.conf.sample

Zakarot:
Create sample config for Hoarder

https://github.com/hoarder-app/hoarder


hoarder.subdomain.conf.sample
```diff
@@ -0,0 +1,47 @@
+## Version 2024/07/16
+# make sure that your <container_name> container is named hoarder-web
```
quietsy, line 2: The name of the container and the name of the conf need to align for the auto-proxy mod, I'd suggest going with `hoarder` instead of `hoarder-web` since the official compose doesn't have a container name and the user will have to set one either way.

PR #774, opened 2025-08-07
Fixing jdownloader websockets

therobbiedavis:

## Description
Websockets were failing when trying to connect via reverse proxy

Fixes websocket failing to connect.

## How Has This Been Tested?
Once change was applied websockets were able to connect.

## Source / References
https://github.com/jlesage/docker-jdownloader-2/issues/265#issuecomment-3039012571

### DIFF
--- a/.github/workflows/check_samples.yml
+++ b/.github/workflows/check_samples.yml
@@ -12,7 +12,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v4.1.7
+        uses: actions/checkout@v4.2.2
 
       - name: Check Allowed File Names
         run: |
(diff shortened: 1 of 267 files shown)


drizuid:
> Thanks but have you tested this?
> 
> I do realize what we had before was incorrect, but the proposed fix also looks incorrect.
> 
> The reason is, when you use a variable in the proxy_pass line, nginx has this weird behavior where it automatically appends the `location` to the end of the proxy_pass address. Really weird behavior that is not really documented anywhere that I could find. It doesn't do that when you use an actual address like an ip or a dns hostname.

i did not test (i dont use jdownloader), but i did look through jdownloaders forums and saw similar suggestions for npm and pure nginx.

aptalca:
Right, unfortunately most of those guides use an IP or a dns hostname so they don't hit that weird behavior. We use vars to make sure nginx doesn't complain and refuse to start when a proxied service is not up or reachable. Another weird nginx behavior.

That's why when we have a separate server block for api endpoints that bypass auth, we don't add `/api` to the proxy_pass line: https://github.com/linuxserver/reverse-proxy-confs/blob/master/bazarr.subdomain.conf.sample#L61

But if it works, I'm ok with merging it. @therobbiedavis can you update the date on the first line as well so it shows up in SWAG log as a conf that needs an update?

Thanks

PR #814, opened 2026-08-11
Joplin: Remove authentication for /api endpoint

NaruZosa:

## Description
Removed authentication for the /api endpoint, as this breaks communication via the client application which handles authentication natively, or for the /api/ping endpoint which shouldn't be protected.

Fixes communication with the Joplin clients when using authentication such as authelia or tinyauth

## How Has This Been Tested?

Used template on my setup with tinyauth enabled
Confirmed /api/ping is reachable without auth
confirmed mobile and desktop apps can communicate with the joplin server
Confirmed auth is required for accessing non-api endpoints, such as the root


NaruZosa:
Not stale.

